PR #6, opened 2020-02-23
use latest version of gql_build

smkhalsa:


lib/src/req_builder/req_code_builder.dart
```diff
@@ -15,26 +14,32 @@ Library buildReqLibrary(
...
             dataDocUrl,
           ),
         ),
     );
 
-List<Class> _buildOperationReqClasses(DocumentNode doc, DocumentNode schema,
-        String opDocUrl, String varDocUrl, String dataDocUrl) =>
+List<Class> _buildOperationReqClasses(
+  DocumentNode doc,
+  String opDocUrl,
```
klavs: You should now be able to use the URI fragments to achieve the same.

Issue #9, opened 2020-02-24
Consider removing Query Widget


smkhalsa:
I'd like to keep the API footprint as minimal as possible without sacrificing core functionality.

The `Query` widget is a very light wrapper around `StreamBuilder`. `StreamBuilder` [re-subscribes](https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/widgets/async.dart#L117) whenever the stream changes. The only value that `Query` is adding currently over using `StreamBuilder` directly is that it prevents re-subscribing whenever the parent widget rebuilds. Since `QueryRequest` doesn't currently override "==", if the `QueryRequest` is created within the build method of the parent widget, it will have a new identity on every build and will cause `StreamBuilder` to re-subscribe.

By simply overriding "==" in `QueryRequest` (either directly or by implementing `built_value`) we should be able to just use `StreamBuilder` directly and remove the `Query` Widget altogether.

This should reduce the API footprint and make it more transparent to the user how the library works.

iscriptology:
I think it conceptually wrong to develop a flutter library with `make it more transparent to the user how the library works.` in mind.

IMO (almost)nobody want to know how things work, people (and software!) want abstractions that work.

smkhalsa:
Maybe I didn't express my primary point clearly: I don't believe the query
widget actually provides a useful abstraction over using StreamBuilder
directly and it adds bloat to the API.

smkhalsa:
Here's a comparison of how you'd use the `Query` widget vs the native `StreamBuilder` widget.

With `Query`

```dart
Widget build(BuildContext context) {
  return Query(
    client: client,
    queryRequest: MyQuery(),
    builder: (BuildContext context, QueryResponse<$MyQuery> response) {
      final data = response?.data;
      /// do something with data
    });
}
```

With `StreamBuilder`

```dart
Widget build(BuildContext context) {
  return StreamBuilder(
    stream: client.responseStream(MyQuery()),
    builder: (BuildContext context, AsyncSnapshot<QueryResponse<$MyQuery>> snapshot) {
      final data = snapshot.data?.data;
      /// do something with data
    },
  );
}
```

iscriptology:
`AsyncSnapshot<QueryResponse<$MyQuery>> snapshot` is a really not pleasant line to write.
also if you want to implement polling, caching, and more features a dedicated widget could be the place to go.

smkhalsa:
[...]

PR #10, opened 2020-02-25
Remove query

smkhalsa:
per #9, this PR removes the `Query` widget in favor of using `StreamBuilder` directly. This change narrows the API without losing core functionality and allows us to remove the `Flutter` dependency.

`QueryRequest` also now implements `Request` from `gql_exec`.


smkhalsa:
> As a thought experiment (and a bit of a stretch), what if there was no `QueryRequest` and it's config was just a `class FerryContext extends ContextEntry`?

Interesting idea!

In this case, [queryController](https://github.com/gql-dart/ferry/blob/master/lib/src/client/client.dart#L29) could simply be a stream of `Query` instead of `QueryRequest<T>`. I'm not sure how we'd deal with `QueryRequest<T>` generics since we [filter by type](https://github.com/gql-dart/ferry/blob/master/lib/src/client/client.dart#L42) in responseStream.

smkhalsa:
After digging deeper on this, I'm going to hold off on removing the Query Widget for now.

Even with the equality override on `QueryRequest`, the `StreamBuilder` will constantly rebuild since calling `client.responseStream(MyQuery())` will still return a unique `Stream` object each time.

I've thought about some workarounds, but I think it's easier just to maintain the `Query` widget for now.

Some of the other changes in this PR have been incorporated into https://github.com/gql-dart/ferry/pull/11.

Issue #8, opened 2020-02-24
Improve error handling


smkhalsa:
We should explore adding an [error policy](https://www.apollographql.com/docs/react/data/error-handling/#error-policies) and [error link](https://www.apollographql.com/docs/link/links/error/), similar to apollo.

PR #23, opened 2020-04-22
Handle network errors

smkhalsa:
Addresses some issues with errors (#8), including catching network errors and including them in `QueryResponse`s.


lib/src/client/query_response.dart
```diff
@@ -1,42 +1,83 @@
...
+import "package:collection/collection.dart";
 
 import './query_request.dart';
 
+enum ResponseSource {
+  /// A placeholder response source which can be used when waiting for another source
+  None,
+
+  /// Response originated from the client [Link]
+  Network,
```
klavs, line 12: I wonder if this should be renamed to `Link`. Because a link may create a response without going through the network.

lib/src/client/query_response.dart
```diff
@@ -1,42 +1,83 @@
...
 
   /// The typed data of this response.
   final T data;
 
   /// The list of errors in this response.
-  final List<GraphQLError> errors;
+  final List<GraphQLError> graphqlErrors;
+
+  /// Any error returned by [Link]
+  final Object networkError;
```
klavs, line 38: We do have `LinkException` (https://github.com/gql-dart/gql/blob/master/gql_link/lib/src/exceptions.dart).

But it might not be good enough if the link is not well implemented.

lib/src/client/query_response.dart
smkhalsa: That's a great point. I'll change it.

lib/src/client/query_response.dart
smkhalsa: Per your suggestion above, I think I will rename this to `linkException`. Even though `networkError` is used elsewhere in the graphql ecosystem, I think `linkException` more correctly captures what it is.

Regarding the type, I think we could handle this in one of two ways.

1. Leave it as an `Object` type and the user can type-check in their code when handling the error
2. Change the type to `LinkException` and only emit a `QueryResponse` when exceptions are of that type. All other exceptions would be handled as error events in the stream.

smkhalsa: @klavs I'm leaning towards # 2. Thoughts?

klavs: Yeah, I agree. Even if we find we want to catch those other errors as well, it'll be quite easy to fix it by creating a subclass of `LinkException`.

PR #32, opened 2020-05-06
Fix query response has errors

TarekkMA:
fixed the condition for QueryResponse.hasErrors and added tests for it.

also added the build folder to the gitignore file since it introduced unwanted files which increased the repo size

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ pubspec.lock
 # Documentation
 doc/api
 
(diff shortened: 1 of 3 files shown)


TarekkMA:
thank you for the great library.

can you delete the build/ dir from the repo it adds about 25 MB to the size

smkhalsa:
yes, done

PR #34, opened 2020-05-07
add test, update pubspec in example, add idea to gitignor

awaik:
1. made a test for writing to the cache
2. added to .gitignor idea folder
3. updated dependencies in the example


awaik:
I've deleted the previous PR and made this with branch and current updates.

PR #35, opened 2020-05-10
Restructure into monorepo

smkhalsa:


.github/workflows/ferry_flutter_tests.yml
```diff
@@ -0,0 +1,25 @@
+name: Ferry Flutter Tests
+
+on:
+  push:
+    paths:
+      - "ferry_flutter/*"
```
klavs, line 6: You might want to run these tests when `ferry` changes as it may break some contracts `ferry_flutter` relies on.

These kinds of dependencies are the reason I wrote and use `multipack`.

.github/workflows/ferry_hive_store_tests.yml
```diff
@@ -0,0 +1,25 @@
+name: Ferry Hive Store Tests
```
klavs, line 1: There's that `strategy.matrix` to run similar jobs in parallel (https://github.com/gql-dart/gql/blob/331dd62f33f14efb30d8ad24efc00746fbf827ce/.github/workflows/dart.yml#L7)

README.md
```diff
@@ -1 +1,87 @@
...
+
+[github-watch-badge]: https://img.shields.io/github/watchers/gql-dart/ferry.svg?style=for-the-badge&logo=github&logoColor=ffffff
+[github-watch-link]: https://github.com/gql-dart/ferry/watchers
+[github-star-badge]: https://img.shields.io/github/stars/gql-dart/ferry.svg?style=for-the-badge&logo=github&logoColor=ffffff
+[github-star-link]: https://github.com/gql-dart/ferry/stargazers
+[github-forks-badge]: https://img.shields.io/github/forks/gql-dart/ferry.svg?style=for-the-badge&logo=github&logoColor=ffffff
+[github-forks-link]: https://github.com/gql-dart/ferry/network/members
+
+[discord-badge]: https://img.shields.io/discord/559455668810153989.svg?style=for-the-badge&logo=discord&logoColor=ffffff
+[discord-link]: https://discord.gg/NryjpVa
```
klavs, line 26: I think you should be able to create an invite URL from discord so that it shows your name instead of mine when somebody clicks on that link.

smkhalsa: good call. Done

.github/workflows/ferry_hive_store_tests.yml
smkhalsa: Interesting. Do you know if this allows for deeply nested paths? I nested the pokemon example in `/ferry/example/pokemon-explorer`.

I didn't want to muddy the top-level directory with examples, especially as we continue to add more. Also, since I added a `README.md` to the `/ferry/example` directory, it will show up on pub.dev.

I'll have to explore this further.

.github/workflows/ferry_flutter_tests.yml
smkhalsa: Yes, great point. I'll try to port over your workflow from gql

.github/workflows/ferry_hive_store_tests.yml
klavs: If you use `multipack` you can refer to packages by name instead of directory.

klavs: I use top-level examples because to me it's a more natural place to put examples with dep on more than one of the packages.

.github/workflows/ferry_flutter_tests.yml
smkhalsa: @klavs I'm just digging into this now. Should I be able to just copy https://github.com/gql-dart/gql/blob/331dd62f33f14efb30d8ad24efc00746fbf827ce/.github/workflows/dart.yml and replace the package names?

.github/workflows/ferry_hive_store_tests.yml
smkhalsa: Yea, thinking about this more, that makes more sense. I'll move examples to a top level folder which should also make them more discoverable.

.github/workflows/ferry_flutter_tests.yml
klavs: Overall, yes. Note that there are two jobs. One for packages and the other one for examples (to compare codegen results).
You should take the latest version (https://github.com/gql-dart/gql/blob/master/.github/workflows/dart.yml) which uses the published `multipack`.

smkhalsa: Awesome. The auto Alpha publishing is pretty slick too. I might try to port that over as well.

klavs: I'm currently trying to figure out how to fully and correctly automate the release process. I am not sure the alpha publishing is stable and actually usable by the package end-users, but at least there's a chance.

smkhalsa: What's the recommended way to handle packages that depend on flutter? The workflow runners are getting `ProcessException: No such file or directory` when trying to run `flutter get packages`.

klavs: Yeah... I guess I never figured this out. The problem is that `google/dart` docker image does not include flutter.

Maybe something like this would work: https://hub.docker.com/r/cirrusci/flutter/tags

klavs: Or this one: https://github.com/marketplace/actions/flutter-action

smkhalsa: Hmm. Neither of those worked out of the box for me. I've just removed the flutter packages from workflows for now.

PR #36, opened 2020-05-26
fixed dependencies in ferry_flutter & example + added line to the git…

awaik:
Hi!

1. The example didn't work - fixed. The path for packages changed from remote to local.
2. Fixed dependencies in ferry_flutter.
3. Added exception for .idea to the gitignor.

I can fix other dependencies in the other packages, if you don't mind.

### DIFF
--- a/examples/pokemon_explorer/pubspec.yaml
+++ b/examples/pokemon_explorer/pubspec.yaml
@@ -4,19 +4,23 @@ environment:
   sdk: '>=2.7.0 <3.0.0'
 dependencies: 
   gql_link: ^0.3.0
-  gql_http_link: ^0.2.2
+  gql_http_link: ^0.2.9
   ferry: ^0.4.0
-  ferry_generator: ^0.1.0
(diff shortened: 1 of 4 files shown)


awaik:
Please note, without this PR package ferry_flutter doesn't work. So, please, publish it on pub.dev

examples/pokemon_explorer/pubspec.yaml
```diff
@@ -4,19 +4,23 @@ environment:
...
   gql_link: ^0.3.0
-  gql_http_link: ^0.2.2
+  gql_http_link: ^0.2.9
   ferry: ^0.4.0
-  ferry_generator: ^0.1.0
-  ferry_hive_store: ^0.1.0
-  hive: ^1.3.0
-  hive_flutter: ^0.3.0+1
-  get_it: ^3.1.0
+  ferry_generator:
```
smkhalsa, line 9: Since the example is meant to be a template, we shouldn't use path references in pubspec.

PR #40, opened 2020-06-14
Fix that fragment data was not merged after denormalization

### LINKED ISSUE #39: Denormalization does not merge data of expanded fragments
There is an issue with the `denormalizeNode` function in the `normalize` package.

It denormalizes all fragments in a query one by one. The values of the fields requested by a fragment are then stored in a `result` map under the key of the queried field. If the same field is queried again by another fragment, the value stored before is replaced.

This is problematic if multiple fragments query an object, but query different fields of that object. In that case only the object data of the last processed fragment is present in the result.

Here is an example. Consider the following query and fragments.

```graphql
fragment MovieStarComponent on MovieType {
	__typename
	id
	name
	starRole {
		__typename
		id
		name
		age
		haircolor
	}
}

fragment MovieTitleComponent on MovieType {
	__typename
	id
	name
	starRole {
		__typename
		id
		name
	}
}

query MovieByID($movieID: ID!) {
	movie(id: $movieID) {
		... MovieStarComponent
		... MovieTitleComponent
	}
}
```

Running `denormalize` on this query will result in the following map:

```
{
	movie {
		__typename: MovieType,
		id: 1,
		name: Rocky,
		starRole {
			__typename: ActorType,
			id: 1,
			name: Sylvester Stallone,
		}

	}
}
```

Notice that the `starRole` objects `age` and `haircolor` fields are not returned, even though they are queried by the `MovieStarComponent` fragment.

Here the `MovieStarComponent` fragment is processed first. The `result` map now contains the queried movies `starRole` object and all fields queried by the `MovieStarComponent` fragment.

Next the `MovieTitleComponent` fragment is processed. It queries the same  `starRole` object but less fields than the `MovieStarComponent` fragment.

Due to the bug described before, the data queried by the `MovieStarComponent` fragment is replaced with the data of the `MovieTitleComponent` fragment.

Changing the order of the fragments in this particular query would "fix" the issue in this situation.

#### Solution

One solution to fix this problem would be to deepmerge the data of denormalized fragments on the same level, instead of just keeping the data of the last processed fragment. I will create an PR in a minute with such changes.

DoctorJohn:
This PR fixes #39 and adds an test demonstrating the issue.

### DIFF
--- a/normalize/lib/src/denormalize.dart
+++ b/normalize/lib/src/denormalize.dart
@@ -7,6 +7,7 @@ import './classes/type_policy.dart';
 import './helpers/resolve_root_typename.dart';
 import './classes/add_typename_visitor.dart';
 import './classes/partial_data_exception.dart';
+import './helpers/deep_merge.dart';
 
 /// Denormalizes data for a given query
 ///
@@ -113,11 +114,17 @@ Map<String, dynamic> denormalize({
           if (!denormalizedData.containsKey(fieldName)) {
             if (!returnPartialData) throw PartialDataException();
           } else {
-            result[selection.alias?.value ?? selection.name.value] =
-                denormalizeNode(
+            final key = selection.alias?.value ?? selection.name.value;
+            final existingValue = result[key];
+            final newValue = denormalizeNode(
               node: selection,
               dataForNode: denormalizedData[fieldName],
             );
+            if (existingValue is Map && newValue is Map) {
+              result[key] = deepMerge(existingValue, newValue);
+            } else {
+              result[key] = newValue;
+            }
           }
           return result;
         },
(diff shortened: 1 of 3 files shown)


smkhalsa:
@DoctorJohn. Great catch, and thanks for the PR.

Rather than deeply merging data, I think it'd be better to deeply merge the selection sets so that the data only gets copied once. 

Since we already have the `expandFragments` helper function, we can simply deeply merge field nodes. This also means we don't have to duplicate the logic in `denormalizeFragment.dart`.

I've implemented an alternative fix in #41 using this approach. Your test is passing. Please give it a try and let me know what you think.

DoctorJohn:
#41 Looks and works great. Unfortunately my knowledge of the ferry code base wasn't broud enough to came of with an solution that elegant. Thanks for taking the time to implement an alternative fix.

smkhalsa:
closing in favor of #41

PR #41, opened 2020-06-15
Fix bug with fragments on same object with different fields

### LINKED ISSUE #39: Denormalization does not merge data of expanded fragments

smkhalsa:
fixes #39. An alternative to #40.


iscriptology:
Publish ETA?

smkhalsa:
It's already published

smkhalsa:
https://pub.dev/packages/normalize

DoctorJohn:
ferry and ferry_cache list normalize ^0.1.1 in their dependencies. Shouldn't we update those?

iscriptology:
I meant under ferry

smkhalsa:
Done. New versions published for `ferry` and `ferry_client`.

PR #43, opened 2020-06-15
Remove the type check for FragmentSpreadNodes in expandFragment

### LINKED ISSUE #42: Expanding fragments defined on interfaces
The `expandFragments` function in the normalize package only includes fragments if their type name matches the data objects type name. This is a problem if the fragment is defined on an interface. GraphQL servers return concrete typed data even when a query contains an fragment that is defined on an interface.

As an example consider the following type, interface and query (slightly inspired by the github graphql api):

```graphql
interface Actor {
    id: ID!
    avatarUrl: String!
}

type User implements Actor {
    id: ID!
    avatarUrl: String!
    name: String!
}
```

```graphql
query Users {
    users {
      ... ActorInfo
    }
}

fragment ActorInfo on Actor {
    __typename
    id
    avatarUrl
}
```

A response to this query could look something like this:

```json
"users": [
    {
        "__typename": "User",
         "id": "1",
         "avatarUrl": "http://example.com/avatar.png"
    }
]
```

Running `normalize` on it should return the following normalized map:

```json
"Query": {
    "users": [
        {
            "$ref": "UserType:1"
        },
    ],
},
"User:1": {
    "__typename": "User",
    "id": "1",
    "avatarUrl": "http://example.com/avatar.png",
}
```

In `expandFragments` the normalized map (variable `data`) is used to perform the following check:

```dart
// Only include this fragment if the type name matches
if (fragment.typeCondition.on.name.value == data["__typename"]) {
    // ...
}
```

The fragments type name will be `Actor` but the datas (concrete) type will be `User`, thus the fragment wont't be included.

DoctorJohn:
Thats what I tried to fix #42. While a quick test was working, other tests are apparently failing. Need to take another look.

### DIFF
--- a/normalize/lib/src/helpers/expand_fragments.dart
+++ b/normalize/lib/src/helpers/expand_fragments.dart
@@ -23,14 +23,11 @@ List<FieldNode> expandFragments({
       }
     } else if (selectionNode is FragmentSpreadNode) {
       final fragment = fragmentMap[selectionNode.name.value];
-      // Only include this fragment if the type name matches
-      if (fragment.typeCondition.on.name.value == data["__typename"]) {
-        fieldNodes.addAll(expandFragments(
-          data: data,
-          selectionSet: fragment.selectionSet,
-          fragmentMap: fragmentMap,
-        ));
-      }
+      fieldNodes.addAll(expandFragments(
+        data: data,
+        selectionSet: fragment.selectionSet,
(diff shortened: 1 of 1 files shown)


smkhalsa:
Thank you!

smkhalsa:
The CI was failing due to a mistake in a test, but I'll take care of that.

PR #48, opened 2020-08-01
Fix for rebuilding outside StreamBuilder

aaahrens:
I've detailed the following in an example repository here https://github.com/aaahrens/example_bug.

While wrapping the Query widget with a StreamBuilder it wont update children that use out of scope variables from the builder.

I think this might have something to do with GC or how states are diffed.

Using the builder from the widget and not the state fixes this problem. 

Also: my flutter channel is on dev.


smkhalsa:
@aaahrens this is marked as WIP. Is there anything else needed before I merge it?

PR #52, opened 2020-08-11
Notify Link when stream isn't being listened to anymore

### LINKED ISSUE #51: Link does not get notified when stream isn't listened anymore
Hi,

I think it would be nice if a custom Link would get notified when the Stream it is returning in it's `request` method isn't being listened to anymore.
This would allow for the Link to cancel a Subscription automatically once no one is listening anymore, or to clean up other resources.

Currently, this is not the case. 

This can be shown in this test case, which will time out:

```dart
import 'dart:async';

import 'package:ferry/ferry.dart';
import 'package:gql_exec/gql_exec.dart';
import 'package:gql_exec/src/response.dart';
import 'package:test/test.dart';

import './graphql/all_pokemon.req.gql.dart';


class _StreamCancelTestLink extends Link {
  final Completer hasCanceledStreamCompleter = Completer();

  @override
  Stream<Response> request(Request request, [forward]) async* {
    StreamController<Response> controller =
        StreamController(onCancel: () => print("oncancel controller"));
    StreamSubscription sub;
    try {
      //simulate events coming in, e.g. from a websocket
      sub =
          Stream.periodic(Duration(microseconds: 1), (_) => Response(data: {}))
              .listen(controller.add);
      //yield* should finish when the client stops listening
      yield* controller.stream;
    } finally {
      //this gets called when the client stops listening to the stream
      sub?.cancel();
      controller.close();

      hasCanceledStreamCompleter.complete();
    }
  }
}

void main() {
  test("Can cleanup", () async {
    final link = _StreamCancelTestLink();

    final client = Client(link: link);
    final req1 = AllPokemon(
        fetchPolicy: FetchPolicy.NetworkOnly, buildVars: (b) => b..first = 3);
    expect(link.hasCanceledStreamCompleter.isCompleted, isFalse);
    StreamSubscription subscription;
    subscription = client.responseStream(req1).skip(1).listen((event) async {
      //cancel subscription after first item
      //this ensures the the link.request() method is entered
      await subscription.cancel();
    });
    // the link should complete its "yield*" statement when the subscription is cancelled
    // and complete its completer
    await link.hasCanceledStreamCompleter.future;

    expect(link.hasCanceledStreamCompleter.isCompleted, isTrue);
  });

}
```

I think with two small changes to the Ferry client we can get this behavior. I will open PRs with these changes shortly.

knaeckeKami:
This is a partial fix for #51.
This will fix #51 in all cases except when the Client uses FetchPolicy.CacheAndNetwork.

For FetchPolicy.CacheAndNetwork, a separate fix is needed. I will suggest a fix for this in a separate PR, because I am not really sure why it happens and not confident in my fix, so maybe someone with a better knowledge of rxdart can find a better solution.

### DIFF
--- a/ferry/lib/src/client.dart
+++ b/ferry/lib/src/client.dart
@@ -160,24 +160,22 @@ class Client {
   /// Fetches the query from the network, mapping the result to a
   /// [QueryResponse].
   Stream<QueryResponse<T>> _networkResponseStream<T>(
-      QueryRequest<T> queryRequest) async* {
+      QueryRequest<T> queryRequest) {
     try {
-      await for (var response in link.request(queryRequest)) {
-        yield QueryResponse(
-          queryRequest: queryRequest,
-          data: (response.data == null || response.data.isEmpty)
-              ? null
-              : queryRequest.parseData(response.data),
-          graphqlErrors: response.errors,
-          dataSource: DataSource.Link,
-        );
-      }
+      return link.request(queryRequest).map((response) => QueryResponse(
+            queryRequest: queryRequest,
+            data: (response.data == null || response.data.isEmpty)
+                ? null
+                : queryRequest.parseData(response.data),
+            graphqlErrors: response.errors,
+            dataSource: DataSource.Link,
+          ));
     } on LinkException catch (e) {
-      yield QueryResponse(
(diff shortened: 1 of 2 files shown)


smkhalsa:
closing in favor of #53

PR #53, opened 2020-08-11
Fix stream cancelling for FetchPolicy.CacheAndNetwork, too

### LINKED ISSUE #51: Link does not get notified when stream isn't listened anymore

knaeckeKami:
This fixes #51 .

This includes the changes in #52, and fixes for `FetchPolicy.cacheAndNetwork` policy, and extends the test for canceling the streams when they are not listened to anymore for all FetchPolicies.

I am not really sure why these fixes to `FetchPolicy.cacheAndNetwork` are necessary, I know too little about rxdart for this. But with wrapping the network stream in a new StreamController and manually canceling it, the canceling of the stream is passed on the Link, too.

There might be a better with rxdart, which is why I put this fix in a separate PR.

### DIFF
--- a/ferry/lib/src/client.dart
+++ b/ferry/lib/src/client.dart
@@ -171,38 +171,30 @@ class Client {
   /// [QueryResponse].
   Stream<QueryResponse<T>> _networkResponseStream<T>(
       QueryRequest<T> queryRequest) {
-    try {
-      return link.request(queryRequest).transform(
-            StreamTransformer.fromHandlers(
-              handleData: (response, sink) => sink.add(
-                QueryResponse(
-                  queryRequest: queryRequest,
-                  data: (response.data == null || response.data.isEmpty)
-                      ? null
-                      : queryRequest.parseData(response.data),
-                  graphqlErrors: response.errors,
-                  dataSource: DataSource.Link,
-                ),
-              ),
-              handleError: (error, stackTrace, sink) => sink.add(
-                QueryResponse<T>(
-                  queryRequest: queryRequest,
-                  linkException: error is LinkException
-                      ? error
-                      : ServerException(
-                          originalException: error, parsedResponse: null),
-                  dataSource: DataSource.Link,
-                ),
+    return link.request(queryRequest).transform(
+          StreamTransformer.fromHandlers(
+            handleData: (response, sink) => sink.add(
+              QueryResponse(
+                queryRequest: queryRequest,
+                data: (response.data == null || response.data.isEmpty)
+                    ? null
+                    : queryRequest.parseData(response.data),
+                graphqlErrors: response.errors,
+                dataSource: DataSource.Link,
               ),
             ),
-          );
-    } on LinkException catch (e) {
-      return Stream.value(QueryResponse(
-        queryRequest: queryRequest,
-        linkException: e,
-        dataSource: DataSource.Link,
-      ));
-    }
+            handleError: (error, stackTrace, sink) => sink.add(
+              QueryResponse<T>(
+                queryRequest: queryRequest,
+                linkException: error is LinkException
+                    ? error
+                    : ServerException(
+                        originalException: error, parsedResponse: null),
+                dataSource: DataSource.Link,
+              ),
+            ),
+          ),
+        );
   }
 
   /// Fetches the query from the cache, mapping the result to a
--- a/ferry/lib/src/client.dart
+++ b/ferry/lib/src/client.dart
@@ -172,14 +172,30 @@ class Client {
   Stream<QueryResponse<T>> _networkResponseStream<T>(
       QueryRequest<T> queryRequest) {
     try {
-      return link.request(queryRequest).map((response) => QueryResponse(
-            queryRequest: queryRequest,
-            data: (response.data == null || response.data.isEmpty)
-                ? null
-                : queryRequest.parseData(response.data),
-            graphqlErrors: response.errors,
-            dataSource: DataSource.Link,
-          ));
+      return link.request(queryRequest).transform(
+            StreamTransformer.fromHandlers(
+              handleData: (response, sink) => sink.add(
+                QueryResponse(
+                  queryRequest: queryRequest,
+                  data: (response.data == null || response.data.isEmpty)
+                      ? null
+                      : queryRequest.parseData(response.data),
+                  graphqlErrors: response.errors,
+                  dataSource: DataSource.Link,
+                ),
+              ),
+              handleError: (error, stackTrace, sink) => sink.add(
+                QueryResponse<T>(
+                  queryRequest: queryRequest,
+                  linkException: error is LinkException
+                      ? error
+                      : ServerException(
+                          originalException: error, parsedResponse: null),
+                  dataSource: DataSource.Link,
+                ),
+              ),
+            ),
+          );
     } on LinkException catch (e) {
       return Stream.value(QueryResponse(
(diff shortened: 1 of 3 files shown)


smkhalsa:
Thanks for the PR!

I remember there was a reason we used async await for the network request, but I'm travelling now and can't dog into it. 

I'll try to take another look next week.

ferry/lib/src/client.dart
```diff
@@ -160,24 +170,22 @@ class Client {
...
-        yield QueryResponse(
-          queryRequest: queryRequest,
-          data: (response.data == null || response.data.isEmpty)
-              ? null
-              : queryRequest.parseData(response.data),
-          graphqlErrors: response.errors,
-          dataSource: DataSource.Link,
-        );
-      }
+      return link.request(queryRequest).map((response) => QueryResponse(
```
smkhalsa, line 175: I believe the reason I originally used async await here is so that data events will continue to be received after an error. For example, we might have a GraphQL Subscription that gives us the following events:

Data  --  Data -- Error -- Data

If we simply map the stream, I believe it will stop listening for data events after the error, but I haven't tested this.

We might be able to solve both issues with a [StreamTransformer](https://api.dart.dev/stable/2.9.1/dart-async/StreamTransformer-class.html).

knaeckeKami: Ah, I see. That's a good point.
However, I don't think this is possible with `await for`.

See the documentation:

> The final three functions are more special. They involve error handling which an await for loop can’t do—the first error reaching the loops will end the loop and its subscription on the stream. There is no recovering from that. You can use handleError() to remove errors from a stream before using it in an await for loop.
https://dart.dev/tutorials/language/streams#modify-stream-methods

and this dartpad:
https://dartpad.dartlang.org/f194fdda982f703da60d7d237f0c3d76

But in any case, the behaviour with .map() is wrong, it would emit an error when the original implementation would not. I think a StreamTranformer similar to the one in the linked dartpad would be the correct behaviour.

ferry/lib/src/client.dart
```diff
@@ -172,14 +172,30 @@ class Client {
   Stream<QueryResponse<T>> _networkResponseStream<T>(
       QueryRequest<T> queryRequest) {
     try {
```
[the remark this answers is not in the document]
smkhalsa: Otherwise looks good. Thanks!

ferry/lib/src/client.dart
```diff
@@ -118,19 +118,29 @@ class Client {
       case FetchPolicy.CacheAndNetwork:
         {
           final responseStreamFromNetwork =
-              _optimisticNetworkResponseStream(queryRequest).shareValue();
+              _optimisticNetworkResponseStream(queryRequest);
+
+          final StreamController<QueryResponse<T>> controller =
```
smkhalsa, line 123: I'm not sure why wrapping the stream in a new StreamController is necessary either.

Is it still necessary with the changes below using StreamController?

ferry/lib/src/client.dart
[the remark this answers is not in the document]
knaeckeKami: The difference is that the link might throw a synchronous exception. The question is, do we want to convert this exception to a stream with an error event or if we should propagate this exception? I'm fine with both, if you think that throwing an exception is better, I'll get rid of the try/catch block.

ferry/lib/src/client.dart
knaeckeKami: Yes, if I omit the wrapping, the test for propagating the cancel event to the link fails on the Fetchpolicy CacheAndNetwork (which is the default for Subscriptions, to that's kinda bad). 

However, I'm not sure why this is necessary. :/

ferry/lib/src/client.dart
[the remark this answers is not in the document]
smkhalsa: It seems that this would only throw if there was an internal exception within the `Link.request` implementation. In that case, I think it'd be better to propagate the exception.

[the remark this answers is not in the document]
knaeckeKami: Ok, I removed the try/catch block.
I also needed to change a test to match the new behavior, and I added a new test.
These tests verify the following behavior:

 - if the link throws an exception, the Stream returned by the client emits an error event
 - if the link returns a Stream that emits an error event, the client wraps this error event into a QueryResponse

ferry/test/error_test.dart
```diff
@@ -65,7 +64,69 @@ void main() {
...
+        options: ClientOptions(addTypename: false),
+      );
+
+      expect(client.responseStream(allPokemonReq),
+          emitsInOrder([isA<QueryResponse>(), isA<QueryResponse>()]));
+    });
+  });
 }
+
+abstract class Mockable {
```
smkhalsa, line 128: Did you intend to leave the Mockable & MyMock classes in here?

smkhalsa (commented):
I'd like to further investigate the FetchPolicy.CacheAndNetwork code at some point to better understand what's going on, but I think these changes are fine for now. There may be a way to refactor that doesn't require the use of `shareValue` at all. 

Thanks again for the PR!

ferry/test/error_test.dart
knaeckeKami: Ah, my bad. Of course not.

smkhalsa (approved):
LGTM, although it's still a draft PR so it can't yet be merged.

PR #82, opened 2020-09-10
Copy list of Plugins from defaultPlugins when instantiating a client

knaeckeKami:
Currently, the reference to the list of `plugins` of a client is a reference to the global `defaultPlugins` variable by default.
This means, mutating this list also mutates the `defaultPlugins` variable.

This is probably unexpected behavior for most users.

I would assume, that if I created 2 clients and mutate the list of plugins of one client, that this has no effect on the second client.

This PR changes this behavior, the client now copies its `plugins` list from `defaultPlugins` and adds a test that verifies this.


smkhalsa (approved):
Awesome! Thanks for the PR.

PR #83, opened 2020-09-10
Don't add __typename to Subscriptions

### LINKED ISSUE #81: prevent subscription from adding __typename
As per the GraphQL spec, subscriptions must include one and only one root field [including introspection fields](https://spec.graphql.org/June2018/#example-2353b).

I have a `MessageReceived` subscription with one root field:
```graphql
subscription MessageReceived {
    messageReceived {
        prevMessageTimestamp
        message {
            ...on TextMessage {
                ...FullTextMessage
            }
            ...on RecordingMessage {
                ...FullRecordingMessage
            }
            ...on TypingMessage {
                ...FullTypingMessage
            }
            ...on PresenceMessage {
                ...FullPresenceMessage
            }
        }
    }
}
```
 and for some reason I get `Subscription "MessageReceived" must select only one top level field.` error message.

I hight suspect that addTypename functionality automatically adds __typename to my subscription and therefore invalidating it.

1) Is there a way to prevent a certain request from adding the __typename?
2) In general I think we should prevent addTypename on the root `Subscription` type.

smkhalsa:


iscriptology:
Hey, any planned release soon?

PR #96, opened 2020-10-07
Refactor to use TypedLink Architecture

### LINKED ISSUE #85: Refactor pagination to stay subscribed to all response streams
Rather than switchmapping, we should use a streamtransformer to combine streams if the request has an update result callback

smkhalsa:
I haven't been completely satisfied with the ferry architecture, especially with respect to extending the client.

The implimentation of plugins in #67 was an effort to modularize ferry. However, after further thought, I realized that most features in ferry were already following the `Link` pattern (i.e. taking a `Request` and returning a `Stream<Response>`).

Rather than implementing a completely new plugin system, I decided to refactor Ferry's core features as composable `Link`s. However, since `gql_exec`'s `Request` and `Response` [don't use generic types](https://github.com/gql-dart/gql/issues/142), I created an alternative, generic typed version of `gql_exec` called `ferry_exec` which includes a `TypedLink` implementation.

This PR refactors Ferry's core features as `TypedLink`s. In fact, the `Client` object itself is now a `TypedLink`, constructed by composing several other `TypedLink`s.

Since the `Link` pattern is already familiar for developers who work with graphql, I hope this new architecture will make it easier to customize and extend the client.

Another benefit is that features can be more easily isolated for testing.

closes #85


smkhalsa:
We should be able to merge this once https://github.com/ReactiveX/rxdart/pull/515 is merged and published.

PR #100, opened 2020-10-13
fixes #98

### LINKED ISSUE #98: [normalize] Asking the same object in the same operation with less fields results in a failure to cache the operation
I'll explain this bug with an example:
```graphql
type Person implements Node {
    id: ID!
    firstName: String!
    lastName: String!
}

type Message implements Node {
    id: ID!
    content: String!
    sender: Person!
}

type Viewer implements Node {
    id: ID!
    friends: [Person!]!
    latestMessage: Message!
}

type Query {
    viewer: Viewer!
}
```

And let's say in the main screen of my app I want to display a list of friends (first name + last name) and my latest message: (only first name is shown in the sender line of messages on my app)
```graphql
query FetchMainScreen {
    viewer {
        friends {
            id
            firstName
            lastName
        }
        latestMessage {
            id
            content
            sender {
                id
                firstName
            }
        }
    }
}
```

When I fetch this query the first time: `_client.responseStream(GFetchMainScreenReq())` using a cache-enabled fetch policy, it will write the normalized data to the cache. But during this process it will **override** one sender with only firstName attached to it. (because the latestMessage's sender is parsed _after_ the friends list, and it only contains firstName)

So if I run `_client.cache.readQuery(GFetchMainScreenReq())` immediately after it will return null because a `PartialDataException` is raised.

This can't be right, as I just executed this query, and now it's by-design stays _partial_ in the cache.

If, say, I rewrite the query above like that: (latestMessage first, friends after)
```graphql
query FetchMainScreen {
    viewer {
        latestMessage {
            id
            content
            sender {
                id
                firstName
            }
        }
        friends {
            id
            firstName
            lastName
        }
    }
}
```
The cache works as expected (since the overwrite just added a lastName field to the sender)

The source of this issue to my understanding is at `normalize_node.dart:75`

```dart
        if (fieldPolicy?.merge != null) {
          return data
            ..[fieldName] = fieldPolicy.merge(
              existingFieldData,
              fieldData,
              FieldFunctionOptions(
                field: field,
                config: config,
              ),
            );
        }
        // ->>>>>>>>> This overrides the previous person that was normalized in this operation with less complete data.
        return data..[fieldName] = fieldData; 
```

I would suggest implementing a default deepMerge (as indeed happens in across-queries writes to the cache) even if I didn't implement a custom merge in a field policy.

smkhalsa:


smkhalsa:
@iscriptology can you please confirm that this issue is now resolved?

iscriptology:
Getting
```
[VERBOSE-2:shell.cc(210)] Dart Unhandled Exception: NoSuchMethodError: The method 'cancel' was called on null.
Receiver: null
Tried calling: cancel(), stack trace: #0      Object.noSuchMethod (dart:core-patch/object_patch.dart:51:5)
#1      forwardStream.<anonymous closure> (package:rxdart/src/utils/forwarding_stream.dart:42:45)
#2      _StreamController._recordCancel (dart:async/stream_controller.dart:756:38)
#3      _ControllerSubscription._onCancel (dart:async/stream_controller.dart:883:24)
#4      _BufferingStreamSubscription._cancel (dart:async/stream_impl.dart:264:21)
#5      _BufferingStreamSubscription._sendDone (dart:async/stream_impl.dart:414:5)
#6      _BufferingStreamSubscription._close (dart:async/stream_impl.dart:305:7)
#7      _SyncStreamControllerDispatch._sendDone (dart:async/stream_controller.dart:816:19)
#8      _StreamController._closeUnchecked (dart:async/stream_controller.dart:671:7)
#9      _StreamController.close (dart:async/stream_controller.dart:664:5)
#10     _RootZone.runG<…>
[VERBOSE-2:shell.cc(210)] Dart Unhandled Exception: type '_InternalLinkedHashMap<dynamic, dynamic>' is not a subtype of type 'Map<String, dynamic>', stack trace: #0      normalizeNode (package:normalize/src/normalize_node.dart:92:30)
#1      normalizeNode.<anonymous closure> (package:normalize/src/normalize_node.dart:69:27)
#2      ListMixin.fold (dart:collection/list.dart:237:22)
#3      normalizeNode (package:normalize/src/normalize_node.dart:59:19)
#4      normalizeNode.<anonymous closure> (package:normalize/src/normalize_node.dart:26:24)
#5      MappedListIterable.elementAt (dart:_internal/iterable.dart:417:31)
#6      ListIterator.moveNext (dart:_internal/iterable.dart:343:26)
#7      new List.from (dart:core-patch/array_patch.dart:38:29)
#8      new List.of (dart:core-patch/array_patch.dart:68:17)
#9      ListIterable.toList (dart:_internal/iterable.dart:211:44)
#10     normalizeNode (package:normalize/src/normalize_node.dart:33:10)
#11     normalizeNode.<anonymous closure> (package:normalize/s<…>
```
repeatedly

smkhalsa:
@iscriptology I wasn't able to reproduce this error, but can you check if the latest commit resolved it?

iscriptology:
will check soon. I think that if in tests it's a Map and against the store it's an InternalHashMap then the tests don't really reflect the real thing and might not catch errors like this over time.
I wonder why type analysis didn't catch this mismatch.

smkhalsa:
> will check soon. I think that if in tests it's a Map and against the store it's an InternalLinkedHashMap then the tests don't really reflect the real thing and might not catch errors like this over time.
> I wonder why type analysis didn't catch this mismatch.

I believe `_InternalLinkedHashMap` is a valid `Map`. It's the <dynamic, dynamic> to <String, dynamic> conversion that's causing the issue.

iscriptology:
Problem persists. I tried for some time now to reproduce the problem in the test file and I'm not able to.
I suspect that the read method in the tests is configured to be `read: (dataId) => normalizedResult[dataId],` and in the cache implementation it's `read: optimistic ? optimisticReader : (dataId) => store.get(dataId),`
so the store doesn't give the data that was retrieved during the existing operation.
but I'm not sure...

smkhalsa:
@iscriptology what store are you using? I tested the latest version with an app using MemoryStore as well as one using HiveStore, and I couldn't reproduce.

iscriptology:
HiveStore

iscriptology:
at the end of the day I think the normalize_node method shouldn't call write more than once per dataId but rather have its own state and emit the final merged object once per dataId.
Existing data should be derived from the internal state rather then from config.read(), only at the end the cache will merge the results with the store.

smkhalsa:
I don't think that is the root cause of this issue, and it would make things more brittle.

For example, what if we have existing data in the store for a given dataId, then our normalizeNode function needs to write to that dataId twice.

The first time, we'd want to merge in the existing data from the store and add it to the internal state. The second time, we'd want to merge data in from the internal state. 

My guess is that the root issue here is that when HiveStore rehydrates state, it loses the original type arguments for the nested Maps. We should be able to do a type conversion, but without a repro I'm not completely sure where the type mismatch is happening.

smkhalsa:
What version of HiveStore are you using?

iscriptology:
The type mismatch doesn't happen no more. Just the original issue of partial data written to cache. I'll try to reproduce somehow and will send PR

smkhalsa:
> The type mismatch doesn't happen no more. Just the original issue of partial data written to cache. I'll try to reproduce somehow and will send PR

@iscriptology I think we can just get rid of the intermediate write state in the ferry cache.

Ideally, we'd have some sort of transaction API in `ferry_store` that would allow us to read and write before committing. However, I think this should suffice for now.

Please give it a try and confirm that this resolves the issues for you.

iscriptology:
How can we know this will not impact performance significantly in large amounts of data?

smkhalsa:
I don't think it will since the watch stream only gets updated asynchronously. In other words many synchronous puts should only trigger a single new event.

However, if you notice any change in performance, definitely let me know.

At some point, I'd like to add some performance benchmarks to the CI.

iscriptology:
async puts means - again - that the read and write are not in sync... So I don't think it will solve the problem...

smkhalsa:
Read and write are sync. Only watch updates async (default stream behavior)

iscriptology:
I mean that if the store is implemented using filesystem writes, (SystemPreferences for example) so the algorithm will call the store's write many times along the way, instead of aggregating them in-memory and only write to cache at the end.

smkhalsa:
> I mean that if the store is implemented using filesystem writes, (SystemPreferences for example) so the algorithm will call the store's write many times along the way, instead of aggregating them in-memory and only write to cache at the end.

Yes, I agree that there could be some performance optimizations in some Store implementations by batching writes. However, I'm not sure they would be significant, especially for our Store implementations. The tradeoff, of course, is that an intermediate state creates more opportunities for bugs like the one that this PR addresses.

We can optimize if we encounter any performance issues. Until then, I think it's better to simplify.

iscriptology:
Okay, I'll do a final check today/tomorrow and let you know if it fixed the problem

smkhalsa:
@iscriptology were you able to confirm that this PR resolves the issue?

iscriptology:
Sorry I still haven't gotten to it. Will try tomorrow and report

iscriptology:
I have issues testing it right now as it is rebased on the new typed links architecture and my app won't compile.
Sorry I'll try to find time to sit on it in the coming days.

smkhalsa:
@iscriptology should I go ahead and merge this?

iscriptology:
I still haven't checked it, I plan to update my app to the new ferry typed link and then check. Sorry for the delay, kinda stressed on a few TODO items with higher priority. Will update here when I check if it works.

iscriptology:
Hey Sat, I checked and it did fix the issue. 
I hope it won't have too much impact on performance.

Issue #102, opened 2020-10-13
[normalize feature req] validation and safety utils


micimize:
I'm looking at adding some handlers in `graphql/client.dart` to check for / distinguish between malformed response data and cache misconfigurations, and I'm wondering if there's a better way than what I'm currently thinking.

Current thinking:
* `denormalizeOperation` already validates all subtree structure right? So we can create a thin wrapper `validateResponseStructure` that ignores denormalization, or extract the traversal logic into its own helper
* On response, we do a `writeQuery`/`readQuery`. If it returns `null`, we then:
  * throw a `MismatchedDataStructureException` if `validateResponseStructure` is `null`
  * throw a `CacheRoundTripException` otherwise (could be broken down further later).

The main idea is to `validateResponseStructure` independent of the cache. Might be a util better suited for `gql`.

smkhalsa:
I think this would be a great addition.

Would `validateResponseStructure` just return a `bool`? Or would it give some hints as to where the mismatched data is?

If a `bool` is sufficient, we could integrate validation into `normalizeNode` and `denormalizeNode` by default. Then we wouldn't need a separate validation method. We could just wrap `writeQuery` and `readQuery` in a try / catch and catch validation exceptions.

micimize:
Just throwing an exception is sufficient for now – eventually surfacing mismatched paths could be a goal but we can add that to the exception details later as well

smkhalsa:
After thinking about this further, without having access to the schema, we'd really only be able to validate whether the data should be a Map (i.e. the node has a SelectionSet) or a scalar (i.e. the node doesn't have a SelectionSet), and I'm not sure how useful this would be.

So we could either:
1. include the schema as an optional parameter to the normalize functions, and if one is provided, perform a more thorough validation
2. implement validation as a separate function

How do you envision this being used? Would it be called directly by the user? Or are you incorporating validation into the `graphql` client somehow?

micimize:
@smkhalsa we can tell from the `document` which fields the `Map` should have right? Because denormalize has to select the fields. That's mainly what I'm after, is just to validate that the structure is valid given the operation document.

This came up because for whatever reason a user's `result.data` was unexpectedly `null` and they had no idea why. So like, if a server responds to `{ foo, bar }` with `{ foo: 1 }`, `denormalizeOperation(document, data)` will return `null`

smkhalsa:
@micimize ah, ok.

[...]

PR #104, opened 2020-10-14
More robust partial data handling

micimize:
* Adds a `path` to `PartialDataException` to surface invalid paths
* Adds `acceptPartialData=true` to `normalizeFragment` and `normalizeOperation`
  to allow for the rejection of structurally incomplete results (i.e. `{}` != '{ foo: null }')
* `denormalizeFragment` and `denormalizeOperation` now have a `handleException=true` flag which can be disabled to throw rather than return null
* Adds a `validateOperationDataStructure` closing #102 (`handleException=false` by default)
* moved `returnPartialData` to `config.allowPartialData` in `denormalizeNode` as partial data is now a cross cutting concern
* add tests and a few docstrings elsewhere

### DIFF
--- a/normalize/lib/src/config/normalization_config.dart
+++ b/normalize/lib/src/config/normalization_config.dart
@@ -6,12 +6,26 @@ import 'package:normalize/src/utils/resolve_data_id.dart';
 
 class NormalizationConfig {
   final Map<String, dynamic> Function(String dataId) read;
+
+  /// Fragment or operation variables that parameterize the document.
   final Map<String, dynamic> variables;
+
+  /// Mapping of `__typename` to custom [TypePolicy] to use, if any.
   final Map<String, TypePolicy> typePolicies;
   final String referenceKey;
+
+  /// Mapping of fragment names to their definitions for spread resolution.
   final Map<String, FragmentDefinitionNode> fragmentMap;
+
+  /// Resolves a data id [String] for a given object [Map],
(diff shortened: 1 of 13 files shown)


micimize:
Sorry this does more than #102 specified – I realized there's an issue with write/read round-trips always resulting in valid data, because `normalizeOperation` will fill in nulls and then `denormalizeOperation` will accept them.

micimize:
hmm... `validateOperationDataStructure` won't work for purely virtualized fields because it ignores `typePolicies`, but we don't have the ability to do that yet anyways. And once we do we'll be stripping out purely client-side virtualized fields from the query so the validation will stand.

smkhalsa:
> hmm... `validateOperationDataStructure` won't work for purely virtualized fields because it ignores `typePolicies`, but we don't have the ability to do that yet anyways. And once we do we'll be stripping out purely client-side virtualized fields from the query so the validation will stand.

Is there a reason why `validateOperationDataStructure` shouldn't just call `denormalizeOperation` internally rather than re-implementing it? It seems like it'd resolve this issue and keep things more DRY.

micimize:
@smkhalsa could you take a look at this when you get the chance? I kind of ended up making https://github.com/zino-app/graphql-flutter/pull/754 fairly large, and it depends on this

smkhalsa (approved):
LGTM

PR #105, opened 2020-10-19
make test fail by using unsynchronized read/write closures

iscriptology:


iscriptology:
Hey, as we discussed in #100 the cache calls normalizeOperation with unsynced read/write methods. So in this case the deep merge doesnt really help and the data is still missing fields.

iscriptology:
Maybe we should add readFromCache closure, and the regular read must be in sync with the write. `write(val, 1); assert(read(val) == 1)`
Then the deep merge will consider three sources: the currently processed object, its past versions from the current `normalizeOperation` execution and its version in the cache.

PR #113, opened 2020-11-06
remove workaround for rxdart memory leak

knaeckeKami:
On my last PR I included an ugly hack that created a new `StreamController` for each new request `FetchPolicy.CachAndNetwork.`
I did not know why it was required, but the stream did not properly cancel otherwise.

I just read about a fixed memory leak in rxdart 0.25, and immediately thought of this issue.
Turns out, that really was the problem.
With rxdart 0.25.0, this workaround is no longer required (I verified it by downgrading back to 0.24 - link_subscription_cancel_test fails on 0.24, but not on 0.25 when the workaround is removed).

However, now another test failed:
CacheAndNetwork returns correct data now got a network response when it expected a cached response. But this was fixed the adding a delay of 0 milliseconds to the mocked network response (could this be solved in a more elegant way?)

### DIFF
--- a/ferry/lib/src/fetch_policy_typed_link.dart
+++ b/ferry/lib/src/fetch_policy_typed_link.dart
@@ -81,32 +81,21 @@ class FetchPolicyTypedLink extends TypedLink {
                           [_cacheTypedLink.request(operationRequest).skip(1)]),
             );
       case FetchPolicy.CacheAndNetwork:
-        {
-          final responseStreamFromNetwork =
-              _optimisticLinkTypedLink.request(operationRequest);
+        final sharedNetworkStream =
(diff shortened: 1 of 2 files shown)


smkhalsa:
@knaeckeKami nice! Glad to hear it was simply an issue with RxDart, and we aren't crazy.

It seems what's happening is that with a sync result from the network, the `takeUntil` has a result immediately, and the initial stream from the Cache is never listened to.

Since our goal with `FetchPolicy.CacheAndNetwork` is to take the result from the cache until a network result is available, it makes sense that we wouldn't get a cached result if a network result is immediately available. I think we can probably just merge your PR as is with the modified test.

PR #146, opened 2021-02-03
fix FetchPolicy.CacheFirst for multiple link responses

smkhalsa:


janosroden:
@smkhalsa Hi, I tried to setup test cases for both CacheFirst and CacheAndNetwork cases but I just ruined all the tests :) (I have no experience with react).

Anyway, I think the cause of the issue is the `.takeUntil(sharedNetworkStream)` which cancels its listener after the first data.
However didn't find a 'reverse' takeUntil which would emit the other stream until the source stream has data.

Do you have any idea how to keep the functionality and avoid any cancel?

janosroden:
Maybe the mock library can be configured to fail tests which calls streamSubscription.cancel() in any way

smkhalsa:
@janosroden I found and fixed the bug. Please confirm that this is now working correctly.

janosroden:
@smkhalsa Both policies work perfectly, thanks for your help!

PR #144, opened 2021-01-29
Implement Standalone Offline Client

smkhalsa:
@akinsho Per my [comment](https://github.com/gql-dart/ferry/issues/142#issuecomment-769795602), I've created a new package for a standalone offline client. Please add your fixes / improvements to this PR.

- [x] create seperate `ferry_offline_client` package
- [x] move `OfflineMutationTypedLink` from `ferry` package (and remove `ferry` dependency on `hive`)
- [ ] implement sequential ordering
- [ ] implement backoff and retry mechanism
- [ ] ensure mutations [always get executed](https://github.com/gql-dart/ferry/issues/142#issuecomment-768473574) when coming back online (even if there aren't any listeners)

I've created an [`OfflineConfig`](https://github.com/gql-dart/ferry/blob/offline-client/ferry_offline_client/lib/ferry_offline_client.dart#L19) class to which we can add configurations for sequential ordering and retry.

Let me know if you have any thoughts or questions.

PR #147, opened 2021-02-05
Feat: add sequential ordering and initial backoff

akinsho:
This PR adds the ability to process requests in sequence and fixes issues with results not being persisted.

- [x] Test sequential ordering

- [ ] Decide on `linkExceptionHandler` logic
- [ ] Add example

Update: Exponential back off has been descoped to a future PR


akinsho:
@smkhalsa I'm getting some naming clashes with the existing offline link, should I remove that one or namespace things inside of the new client package?

smkhalsa:
@akinsho it's actually already [re(moved)](https://github.com/gql-dart/ferry/pull/144/files#diff-17c43de9f09664fdd7cc682d5ad748fb098f207737cc6bdd9b7908d8b4795f1d).

The issue is probably that since `ferry` is a dependency of the new package, you're not seeing the version of `ferry` from the PR.

I use [`multipack`](https://pub.dev/packages/multipack) to handle this and ensure that the dependencies are all overridden.

Just activate it with:

`pub global activate multipack`

Then from the monorepo root directory, you can run `multipack pubspec override` then `multipack pub get` to override all the local dependencies.

When you're ready to commit your changes, you can run `multipack pubspec clean` to remove the overrides.

ferry_offline_client/lib/src/offline_mutation_typed_link.dart
```diff
@@ -46,29 +58,48 @@ class OfflineMutationTypedLink extends TypedLink {
...
+  void _handleOnConnect() async {
+    // start at one since the first attempt is the first call
+    var count = 1;
+    await retry(
+      _processRequests,
+      maxAttempts: retryAttempts,
+      onRetry: (error) async {
+        count++;
+        if (count == retryAttempts) {
+          await retriesExhaustedHandler(error);
```
akinsho, line 82: Note: Need to check that this is specified before calling it

ferry_offline_client/lib/src/offline_mutation_typed_link.dart
```diff
@@ -46,29 +58,48 @@ class OfflineMutationTypedLink extends TypedLink {
...
 
-        // Run unexecuted mutations
-        requestController.add(req);
-      });
+  Future<void> _processRequests() async {
+    final queue = StreamQueue(Stream.fromIterable(mutationQueueBox.values));
+    while (await queue.hasNext) {
+      final json = await queue.next;
+      final req =
+          serializers.deserialize(json) as OperationRequest<Object, Object>;
```
akinsho, line 93: Note: don't need to type cast in this context

ferry_offline_client/lib/src/offline_mutation_typed_link.dart
```diff
@@ -115,12 +145,21 @@ class OfflineMutationTypedLink extends TypedLink {
       _responseTransformer<TData, TVars>() => StreamTransformer.fromHandlers(
             handleData: (res, sink) {
               // Forward along any responses for operations that aren't Mutations
-              if (_isMutation(res.operationRequest) == false) {
+              // or any responses that are optimistic since these should not
+              // cause mutations to be treated as executed
```
akinsho, line 149: NOTE: Make this comment clearer that we are trying to prevent optimistic responses from clearing the saved responses since they don't actually count as the original being executed

ferry_offline_client/lib/src/offline_mutation_typed_link.dart
```diff
@@ -115,12 +145,21 @@ class OfflineMutationTypedLink extends TypedLink {
...
                 return sink.add(res);
               }
 
-              if (res.linkException is ServerException &&
-                  linkExceptionHandler != null) {
+              // if there are any response error pass them to the exception handler
+              // do not allow any requests that have errors to be removed from
+              // the queue as these need to be retried
+              //
+              // TODO decide if the error should be returned here so it bubbles
```
akinsho, line 159: @smkhalsa could use some advice here. In order for the retry mechanism to work the error has to propagate upwards. If a link exception handler is specified the only way that the error will propagate to `_processRequests` is if it uses `sink.addError(error)`. Tbh I'm not sure what purpose this should serve in this retrying context. In my project I just use it to log the error I don't actually need it, but I'm not sure what a more generally applicable use for it would be? and also it means the consumer would have to know to add it to the `sink` which isn't great. I guess knowing what this is useful for will make it easier to know how to work around it or remove it so that I can ensure the error passes through

smkhalsa:
Have you looked at [`StreamQueue`](https://github.com/dart-lang/async/blob/master/lib/src/stream_queue.dart)? I've found it to be helpful.

Testing Streams in Dart is hard.

ferry_offline_client/lib/src/offline_mutation_typed_link.dart
smkhalsa: @akinsho The intention of the linkExceptionHandler was to allow users to handle different types of exceptions differently. Simply adding the request to the queue wouldn't be appropriate for every type of exception. 

For example, if the exception is related to being offline, we probably DO want to enqueue the request. However, what if the request was received by the server but the server sent a server 500 error? We probably don't want to simply rerun the same mutation.

akinsho: Hmm I not sure how we'd tell which request was which? Is this something a user should decide and so something that goes into the docs that if you want a failed offline mutation to be retried the error needs to be added to the sink or do you think theres some way of knowing here and doing it for the user so they don't need to figure this part out since I think it will confuse a lot of people.

Could there be a new data source or some sort of metadata to specify that a request is from the offline cache and then if so they should be queued up otherwise the link exception handler should handle it

smkhalsa:
@akinsho Can you remind me why the retry logic needs to be coupled with offline mutations? Why not create a separate link for retrying, similar to [Apollo](https://www.apollographql.com/docs/react/api/link/apollo-link-retry/)?

I presume some users will want retry logic for their Queries as well as Mutations.

akinsho:
@smkhalsa it's primarily because this is how I implemented it for my own project based on our needs and my PR is basically just moving that logic across. They don't have to be coupled although I don't how to split it out into a separate link i.e. how to create separate links in ferry and how one like this would work 🤔 .

It also seems like it would expand the scope of this a bit, since that would need it's own set of set of tests and examples? could that be done separately?

akinsho:
@smkhalsa where should the example go? atm the only example in the repo is the pokemon one which is in it's on directory. Should I just create an example flutter project within the `ferry_offline_client` directory?

smkhalsa:
I'd suggest putting it in the examples directory

eivihnd:
With respect and gratitude for all the work that has been put into ferry and this PR, and acknowledging that you are busy people: Are there any plans for further work on this PR?

(I am afraid my skills and knowledge isn't at the level required to contribute with code to this PR, at least atm)

smkhalsa:
@eivihnd this is not high on my priority list, but you should be able to incorporate the ideas into your own custom link pretty easily.

Issue #114, opened 2020-11-07
NoSuchMethodError: invalid member on null: 'loading'


garyrutland:
Hi,

Been using the library few a weeks now but hit a problem and not sure how to debug it.

Using the `Operation` widget and using a similar setup to the example I have started getting the response `NoSuchMethodError: invalid member on null: 'loading'` but can't find a way to debug it.

My code looks something similar to this:
```
final GMyReq myRequest = GMyReq((b) => b
      ..fetchPolicy = FetchPolicy.NetworkOnly
      ..vars.uuid = widget.uuid);

return Operation(
  operationRequest: festivalRequest,
  client: _client,
  builder: (BuildContext context, OperationResponse<GMyData, GMyVars> response) {
    print('RESPONSE');
    print(response);
    print('RESPONSE END');

    if (response.loading)
      return Center(child: CircularProgressIndicator());

    // ...
  }
);
);
```

But the `print` command outputs the following:
```
RESPONSE
Instance of 'OperationResponse<GMyData, GMyVars>'
RESPONSE END
RESPONSE
null
RESPONSE END
```

Is there a way for me to find out why null is being returned?
I've added `gql_error_link` and a `ErrorLink` to my client but it doesn't seem to be getting triggered.

Thanks,
Gary

garyrutland:
I was able to update the client so I could print out the response using the `TransformLink` and compared the response of a failing response to a successful response from the same request and the only difference between the two is when I was using a `Date` scalar.

When a value was returned (i.e. 2020-11-07) everything worked fine, but if `null` was returned I got the above issue.

To clarify, the schema is declared as `Date` and not `Date!`.
Is it possible this is a bug?

I might be able to get around the issue by implementing a custom scalar based on the docs.

garyrutland:
So after adding a custom serialiser, while it has made life a bit easier for non-null values any response that has a `null` response doesn't even get to the serialiser.

smkhalsa:
What version of ferry are you using?

Are any exceptions being thrown (including any that are being caught by your ErrorLink?)

garyrutland:
From my `pubspec.yaml` file:

```
  ferry: ^0.8.0
  ferry_flutter: ^0.3.0
  gql_http_link: ^0.3.2
  gql_error_link: ^0.1.1-alpha
  gql_transform_link: ^0.1.5
```

[...]

PR #149, opened 2021-02-09
ferry_flutter: Add error builder to Operation widget

janosroden:
Let developers to handle errors.

### DIFF
--- a/ferry_flutter/CHANGELOG.md
+++ b/ferry_flutter/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.3.3]
+
+- Add error builder to Operation widget
+
 ## [0.3.2]
 
 - update dependencies
--- a/ferry_flutter/lib/src/operation.dart
+++ b/ferry_flutter/lib/src/operation.dart
@@ -1,20 +1,56 @@
 import 'package:flutter/material.dart';
+import 'package:ferry/ferry.dart';
 import 'package:ferry_exec/ferry_exec.dart';
+import 'package:gql_exec/gql_exec.dart';
+
+Widget Function(
+  BuildContext context,
+  Object error,
+  LinkException linkException,
+  List<GraphQLError> graphqlErrors, [
+  OperationRequest operationRequest,
+// ignore: prefer_function_declarations_over_variables
+]) defaultOperationErrorBuilder = (
+  context,
+  error,
+  linkException,
+  graphqlErrors, [
+  operationRequest,
+]) =>
+    Column(
+      children: [
+        if (operationRequest?.operation?.operationName != null)
+          Text(operationRequest.operation.operationName),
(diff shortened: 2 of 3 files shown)


smkhalsa:
@janosroden thanks for your interest in contributing.

Error handling may be handled differently by each application. Simply displaying the errors to users is not always a good idea. Furthermore, developers may want to display errors in different ways (e.g. a dialog). Ferry shouldn't be opinionated about if or how errors are displayed.

If you want to achieve what you've implemented in your PR, you can do so easily by writing a wrapper widget around the existing Operation widget.

janosroden:
@smkhalsa The existing Operation widget doesn't allow to handle stream errors (like in #114).

This pr provides a default builder which can be overwritten in application level and also provides error builder which can be specified for the particular Operation.

So I agree, ferry doesn't have to define how to handle errors but this pr just provides a way to handle errors with a default solution.

If you still think this is overkill then we still can choose to don't call the default handler. That would be a backward compatible change. What do you think?

smkhalsa:
> @smkhalsa The existing Operation widget doesn't allow to handle stream errors (like in #114).

This is a fair point, although I'm not convinced that internal TypedLink errors should be handled at the Widget level.

Ferry already has an `ErrorTypedLink`, although it's not included in the default client TypedLink chain. It's currently untested, but we could look at adding it.

Anything related to displaying errors, however, can be handled as a simple wrapper around the existing widget

janosroden:
If I understand correctly `ErrorTypedLink` can help to catch errors before the stream is created but 'runtime' errors are not its responsibility. However it seems useful, thanks for the hint.

The motivation of this PR was that when I tried the library I found the Operation widget, started using and I got null `response` in the builder (due to an unhandled custom scalar). Then I looked into the code how to access to the original `snapshot.error`. I had to copy because `snapshot.error` is simply ignored. Even the stream is created by the widget so no way to pass a stream which is already prepared with onError.

I'd be happy with the 'simple wrapper' you mentioned but I think it would be useful to access to all errors. Otherwise the builders will look like `if (response == null) return Text('Something went wrong');` :)
I can imagine a link which converts stream errors to `OperationResponse` with link exception.

At least do you agree to don't hide the stream error?

smkhalsa:
Error handling has been admittedly lacking until now. We've discussed in the past how to handle Stream error events (i.e. whether to catch them or let the user catch them), and we decided to let the user catch them.

However, as you mentioned, this has led to some unexpected behavior caused by silently failing TypedLinks. With our TypedLink architecture, it makes sense to add some reasonable error handling defaults to the default Client TypedLink.

I've updated the `ErrorTypedLink` to catch Stream error events and convert them into `OperationResponse`s and have included `ErrorTypedLink` in the default `Client` TypedLink chain.

I've also adjusted `ferry_flutter` to include errors from the StreamBuilder (although there shouldn't be any in the default `Client` due to the above). This is a breaking change since it changes the signature of the builder callback.

Please take a look at the [PR](https://github.com/gql-dart/ferry/pull/151) and let me know if it resolves all issues in your use case.

PR #164, opened 2021-03-23
use melos

smkhalsa:


melos.yaml
```diff
@@ -0,0 +1,13 @@
+name: ferry
+
+packages:
+  - exmaples/**
```
Salakar, line 4: `exmaples/**` -> `examples/**`

smkhalsa: Thanks. Fixed in #172

PR #179, opened 2021-05-06
Remove unnecessary cache read

friebetill:
While going through the timeline of my app, I noticed that every time I read the cache, the cache is read twice. This makes a noticeable effect on the UX of my users, so I am trying to remove the second cache read. (Ferry is on an isolate in my app, so the timelines is only from Ferry code)

#### Timeline with existing code

image
The image above shows the timeline of one query with `CacheFirst`. The query takes about 500ms in total and consists of mainly 6 parts:
- the green block: `Cache.watchQuery (half)` is the timeline from this code in `cache.dart`:
   <details><summary>code</summary>
   <p>
   

   ```dart
   var closed = false;

   final dataChanged = operationDataChangeStream(
     request,
     optimistic,
     optimisticPatchesStream,
     optimisticReader,
     store,
     typePolicies,
     addTypename,
   ).doOnDone(() => closed = true);
   ```
    
   </p>
   </details>

- the orange block `cache.watchQuery readQuery` is the timeline from this code in `cache.dart`:

   <details><summary>code</summary>
   <p>
   
   ```dart
   readQuery(
     request,
     optimistic: optimistic,
   );
   ```
   </p>
   </details>

- the many small turquoise blocks `operationDataChangeStream. dataId map` is the timeline from this code in `operation_data_change_stream.dart`:

   <details><summary>code</summary>
   <p>

   ```dart
   var stream = dataForIdStream(
     dataId,
     store,
     optimistic,
     optimisticPatchesStream,
     optimisticReader,
   );

   /// we only want to subscribe to changes on the operation fields, not the entire operation type)
   if (dataId == rootTypename) {
     stream = stream.map(
       (data) => operationRootData(
         data,
         request,
         typePolicies,
       ),
     );
   }
   ```
   </p>
   </details>

- and the three parts again for the second cache read.

#### Timeline with code from this PR:
image
The image above, shows the same CacheFirst query with the code from this PR. The query takes a total of 230ms, about half the time of the query above.

The two timelines can also be viewed in detail here [dart_timeline.zip](https://github.com/gql-dart/ferry/files/6433942/dart_timeline.zip), preferably via `chrome://tracing/` from a Chromium based browser, as the Flutter timeline viewer is slow.

#### Code
All tests are successful, but I am unsure if it is necessary to close the StreamController manually.

#### How to measure the timeline
To generate the timeline in Flutter Dev, I followed [this instructions](https://medium.com/flutter/profiling-flutter-applications-using-the-timeline-a1a434964af3) and used `Timeline.startSync` and `Timeline.finishSync` in the code to see specific code parts.

#### Another possibility for optimization:
I currently think that there is another problem with the `dataIds` in the file `operation_data_change_stream.dart`. There are too many `dataIds` monitored for a small query. This is also visible in the timeline, these are the many small turquoise lines, which in total take quite a lot of time. I'll take a look at that soon and see if there's anything that can be optimized.

### DIFF
--- a/ferry/lib/src/fetch_policy_typed_link.dart
+++ b/ferry/lib/src/fetch_policy_typed_link.dart
@@ -72,47 +72,35 @@ class FetchPolicyTypedLink extends TypedLink {
       case FetchPolicy.CacheOnly:
         return _cacheTypedLink.request(operationRequest);
       case FetchPolicy.CacheFirst:
-        return _cacheTypedLink.request(operationRequest).take(1).switchMap(
-              (result) => result.data != null
-                  ? _cacheTypedLink.request(operationRequest)
-                  : _optimisticLinkTypedLink
-                      .request(operationRequest)
-                      .doOnData(_removeOptimisticPatch)
-                      .doOnData(_writeToCache)
-                      .switchMap(
-                        (networkResponse) => ConcatStream([
-                          Stream.value(networkResponse),
-                          _cacheTypedLink.request(operationRequest).skip(1),
-                        ]),
-                      ),
-            );
-      case FetchPolicy.CacheAndNetwork:
-        final sharedNetworkStream =
-            _optimisticLinkTypedLink.request(operationRequest).shareValue();
+        final controller = StreamController<OperationResponse<TData, TVars>>();
+        var isInitial = true;
 
         return _cacheTypedLink
             .request(operationRequest)
+            .doOnData((result) {
+              if (isInitial) {
+                isInitial = false;
+                if (result.data == null) {
+                  controller.addStream(_optimisticLinkTypedLink
+                      .request(operationRequest)
+                      .doOnData(_removeOptimisticPatch)
+                      .doOnData(_writeToCache));
+                }
+              }
+            })
             .where((response) => response.data != null)
-            .takeUntil(
-          sharedNetworkStream.doOnData(
-            (_) {
-              /// Temporarily add a listener so that [sharedNetworkStream] doesn't shut down when
-              /// switchMap is updating the stream.
-              final sub = sharedNetworkStream.listen(null);
-              Future.delayed(Duration.zero).then((_) => sub.cancel());
-            },
-          ),
-        ).concatWith([
-          sharedNetworkStream
+            .mergeWith([controller.stream])
+            .distinct();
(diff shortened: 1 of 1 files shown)


ferry/lib/src/fetch_policy_typed_link.dart
```diff
@@ -72,47 +72,35 @@ class FetchPolicyTypedLink extends TypedLink {
...
-                        ]),
-                      ),
-            );
+                      .doOnData(_writeToCache));
+                }
+              }
+            })
+            .where((response) => response.data != null)
+            .mergeWith([controller.stream])
+            .distinct();
```
smkhalsa, line 93: Is the `distinct()` necessary to prevent network data written to the cache from triggering a second identical data event from the cache?

If so, this probably only worked because of [this](https://github.com/gql-dart/ferry/pull/180) oversight since responses with different DataSources should not be equal.

Can you please rebase?

ferry/lib/src/fetch_policy_typed_link.dart
```diff
@@ -72,47 +72,35 @@ class FetchPolicyTypedLink extends TypedLink {
...
               .doOnData(_removeOptimisticPatch)
               .doOnData(_writeToCache)
-              .switchMap(
-                (networkResponse) => ConcatStream([
-                  Stream.value(networkResponse),
-                  _cacheTypedLink.request(operationRequest).skip(1),
-                ]),
-              )
-        ]);
+        ]).distinct();
```
smkhalsa, line 103: Same as above

smkhalsa (changes requested):
Thanks for the PR!

I know that there are a few places that performance could be improved, so this is much appreciated.

> I currently think that there is another problem with the dataIds in the file operation_data_change_stream.dart. There are too many dataIds monitored for a small query. This is also visible in the timeline, these are the many small turquoise lines, which in total take quite a lot of time. I'll take a look at that soon and see if there's anything that can be optimized.

That'd be great.

Out of curiosity, how many records are you reading in this example?

smkhalsa:
Looks like this was automatically closed when the target branch was merged into master. Feel free to re-open targeting master.

friebetill:
Ah, I understand and will hopefully reopen the PR soon. At the moment I'm still looking for the main problem why my app has big memory leaks and is very slow since the Flutter 2 update.

PR #193, opened 2021-05-19
Add support for dataIdFromObject to ferry_cache and swap the order of assignment in denormalize_fragment

### LINKED ISSUE #187: Cache including typename in the dataId always
I have run into some issues when attempting to use `readFragment` on a fragment defined on an interface where it is unable to locate the data in the cache because the cache creates ids that are `$typename:$id` and in `readFragment` (https://github.com/gql-dart/ferry/blob/master/normalize/lib/src/denormalize_fragment.dart#L68) the typename comes from the fragment type, which works great for object spreads but not for interface spreads.

My initial thought would be to allow/encourage a cache mode where the ids are globally unique, as that is the pattern I have seen elsewhere.

What do y'all think? Would a PR that creates such a mode be accepted?

tberman:
This attempts to fix #187 with minimal changes.

I wrote some simple tests, but happy to build out more complex test cases if necessary, normalize seemed to test this functionality pretty well already.

### DIFF
--- a/ferry_cache/lib/src/cache.dart
+++ b/ferry_cache/lib/src/cache.dart
@@ -25,14 +25,13 @@ class Cache {
 
   Cache({
     Store? store,
-    utils.DataIdResolver? dataIdFromObject,
+    this.dataIdFromObject,
     this.typePolicies = const {},
     this.addTypename = true,
     Map<OperationRequest, Map<String, Map<String, dynamic>?>>
         seedOptimisticPatches = const {},
   })  : store = store ?? MemoryStore(),
-        optimisticPatchesStream = BehaviorSubject.seeded(seedOptimisticPatches),
-        dataIdFromObject = dataIdFromObject;
+        optimisticPatchesStream = BehaviorSubject.seeded(seedOptimisticPatches);
 
   /// Reads data for the given [dataId] from the [Store], merging in any data from optimistic patches
   @visibleForTesting
@@ -65,7 +64,7 @@ class Cache {
           store,
           typePolicies,
           addTypename,
-          dataIdFromObject: dataIdFromObject,
+          dataIdFromObject,
         ).doOnDone(() => closed = true);
 
         return NeverStream<TData?>()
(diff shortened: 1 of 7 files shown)


ferry_cache/lib/src/cache.dart
```diff
@@ -24,12 +25,14 @@ class Cache {
 
   Cache({
     Store? store,
+    utils.DataIdResolver? dataIdFromObject,
```
smkhalsa, line 28: You can just use `this.dataIdFromObject` here since the final variable is nullable

ferry_cache/lib/src/fragment_data_change_stream.dart
```diff
@@ -2,22 +2,23 @@ import 'package:normalize/policies.dart';
...
-) {
+    FragmentRequest<TData, TVars> request,
+    bool optimistic,
+    Stream<Map<OperationRequest, Map<String, Map<String, dynamic>?>>?>
+        optimisticPatchesStream,
+    Map<String, dynamic>? Function(String dataId) optimisticReader,
+    Store store,
+    Map<String, TypePolicy> typePolicies,
+    bool addTypename,
+    {DataIdResolver? dataIdFromObject}) {
```
smkhalsa, line 21: For consistency, I'd either make this a positional argument or adjust all the arguments to be optional named arguments. If `fragmentDataChangeStream()` was user-facing, I'd definitely go with the latter, but since this is internal, either is fine.

ferry_cache/lib/src/operation_data_change_stream.dart
```diff
@@ -11,15 +11,15 @@ import './utils/operation_root_data.dart';
...
-) {
+    OperationRequest<TData, TVars> request,
+    bool optimistic,
+    Stream<Map<OperationRequest, Map<String, Map<String, dynamic>?>>?>
+        optimisticPatchesStream,
+    Map<String, dynamic>? Function(String dataId) optimisticReader,
+    Store store,
+    Map<String, TypePolicy> typePolicies,
+    bool addTypename,
+    {DataIdResolver? dataIdFromObject}) {
```
smkhalsa, line 22: Same comment here as `fragmentDataChangeStream()` with respect to using all positional or optional named arguments.

smkhalsa (changes requested):
This looks great. Thank you.

I just had a few nitpicks.

smkhalsa (approved):

PR #199, opened 2021-05-30
feat: Add support for better fragment type resolution

budde377:
This PR adds improved support for fragment resolution by introducing a new map to lookup type relationship

Take the document: 

```graphql
fragment FAuthor on Author {
  __typename
  id
}
fragment FAudience on Audience {
  __typename
  id
  numHands
  isClapping
}
fragment FUser on User {
  id
  __typename
  name
}
query {
  users {
    ...FAuthor
    ...FAudience
    ...FUser
  }
}
```

and the schema 

```graphql

type Query {
  users: [User]
}

interface User { 
  id: ID!
  name: String
}

type Author implements User {
  id: ID!
  name: String
  numBooks: Int
}

type Audience implements User {
  id: ID!
  name: String
  numHands: Int
  isClapping: Boolean
}

```

the current implementation will fail to realize which fragment spreads apply to a given type. E.g. that `FUser` and `FAuthor` should apply to the `Author` type. This leads to partial results on fragments and wrong results on inline fragments.  By adding some introspection in the shape of the `possibleTypeOf` lookup map, which maps object types to possible interfaces and unions, we can resolve this.

If the map is omitted, or if the `__typename` is omitted, this solution will not expand any fragments. This is sound. The current approach expands all fragment spreads and only expands on exact matches for inline fragments, both of which are wrong.

### DIFF
--- a/ferry_cache/lib/src/utils/operation_root_data.dart
+++ b/ferry_cache/lib/src/utils/operation_root_data.dart
@@ -9,13 +9,14 @@ Map<String, dynamic> operationRootData<TData, TVars>(
   Map<String, dynamic>? data,
   OperationRequest<TData, TVars> request,
   Map<String, TypePolicy> typePolicies,
+  Map<String, Set<String>> possibleTypes,
 ) {
   final fieldNames = operationFieldNames(
     request.operation.document,
     request.operation.operationName!,
     (request.vars as dynamic).toJson(),
     typePolicies,
-    {},
+    possibleTypes,
   );
   return {
     for (var fieldName in fieldNames)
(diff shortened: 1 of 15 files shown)


ferry_cache/lib/src/utils/operation_root_data.dart
```diff
@@ -15,6 +15,7 @@ Map<String, dynamic> operationRootData<TData, TVars>(
     request.operation.operationName!,
     (request.vars as dynamic).toJson(),
     typePolicies,
+    {},
```
smkhalsa, line 18: We should include the possible types as an argument to `operationRootData` and pass them in here.

smkhalsa (commented):
Thank you for the PR!

One question: is there an advantage to using a `possibleTypeOf` Map instead of a `possibleTypes` Map? Apollo uses the latter, and it feels more intuitive to define possible types for an interface rather than possible interfaces for a type.

budde377:
> One question: is there an advantage to using a possibleTypeOf Map instead of a possibleTypes Map? Apollo uses the latter, and it feels more intuitive to define possible types for an interface rather than possible interfaces for a type.

I don't see any immediate benefits one way or another. Apollo using the latter is a good argument for going in that direction. I'll rewrite!

smkhalsa (approved):
LGTM

Issue #212, opened 2021-06-29
UpdateCacheHandler not invoke after first subscription response


lotus-x:
Hi, My goal is to update cache when I receive an object using subscription.

My code looks like this,

```dart
final _req = GOnNewObjectReq(
    (b) => b
      ..requestId = 'key'
      ..executeOnListen = false
      ..updateCacheHandlerKey = 'update_cache_key',
  );

// manually register subscription
requestController.add(
      _req.rebuild(
        (b) => b..updateCacheHandlerContext = {'role': role},
      ),
    );

client.request(_req).listen((response) {
      print(response.data);
    });
  }
```
Note that I'm executing subscription later when I have the role for update cache context.

But the issue is when the first response arrive, the CacheHandler has been called, and for the next responses from the server its not.

Only the CacheHandler has the issue, subscription stream get the response.

Is this a bug or I'm missing something?

smkhalsa:
This looks like a bug. It'd be helpful if you could add a failing test though.

lotus-x:
```dart
void main() {
  test('Ferry Subscription', () async {
    // request
    final _req = GSubscriptionReq(
      (b) => b..updateCacheHandlerKey = 'cache_handler',
    );
    // init client
    final client = Client(
      link: _SocketLink(),
      cache: Cache(),
      updateCacheHandlers: {
        'cache_handler': (proxy, response) {
          print('cache invoked');
        },
      },
    );
    await for (final response in client.request(_req)) {
      print(response.data);
    }
  });
}

class _SocketLink extends Link {
  @override
  Stream<Response> request(Request request, [NextLink? forward]) {
    final StreamController<Response> controller = StreamController();
    Stream.periodic(
      const Duration(seconds: 2),
      (_) => const Response(data: {}),
    ).listen(controller.add);
    return controller.stream;
  }
}
```

'cache invoked' only print one time

lotus-x:
I think its intentional to work like this.

In the https://github.com/gql-dart/ferry/blob/master/ferry/lib/src/update_cache_typed_link.dart

```dart
switch (res.dataSource) {
        case DataSource.Optimistic:
          {
            handler(proxy, res);
            runNonOptimistically = false;
            return;
          }
        case DataSource.Link:
        case DataSource.Cache:
          {
            if (runNonOptimistically == false) {
              handler(proxy, res);
              runNonOptimistically = true;
            }
            return;
          }
        case DataSource.None:
          return;
      }
```

[...]

PR #214, opened 2021-06-30
Fixed issue with update cache typed link

lotus-x:
When using a link that emits sequential responses for a single request (like graphql subscription) it only invoke the cache handler for the first response only. So this is a fix for the issue.

`runNonOptimistically` should be true default for the common requests. For a request with `DataSource.Link` it should be marked as `false` then again it should be `true` after update the cache.

### DIFF
--- a/ferry/lib/src/update_cache_typed_link.dart
+++ b/ferry/lib/src/update_cache_typed_link.dart
@@ -124,7 +124,7 @@ class UpdateCacheTypedLink extends TypedLink {
       _updateCacheHandlersStream<TData, TVars>(
     Stream<OperationResponse<TData, TVars>> stream,
   ) {
-    var runNonOptimistically = false;
+    var runNonOptimistically = true;
 
     return stream.doOnData((res) {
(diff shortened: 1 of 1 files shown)


dev-pos:
this should fix #212, right? @lotus-x

smkhalsa:
Thanks for the PR.

As you can see from the failed tests, simply changing the default for `runNonOptimistically` to `true` isn't a valid solution.

Also, it would be great if you could add a new test to the tests folder for the issue you are describing.

lotus-x:
Thanks for reviewing this and I'm sorry to make some quick changes. I will go through the tests and give a reliable solution asap.

Anyway why don't anyone get this issue yet? no one using subscriptions? Am I taking some misleading path?

smkhalsa:
Sounds good. I can't say if anyone else has experienced this yet, but I don't use subscriptions in my project

smkhalsa:
Closing due to inactivity, but please reopen if you're still working on this.

PR #237, opened 2021-08-06
Use a single StreamController in watchQuery (fixes #232)

### LINKED ISSUE #232: Stack overflow after a while if subscription delivers many updates
I noticed if my subscription receives updates frequently stack overflow occurs after a while (~10 mins). Not too scientific, I admit but always reproducible.

I think the reason is the watchQuery call (FetchPolicy -> CacheTypedLink -> Cache) which is recurs*ish*:

```dart
     return NeverStream<TData?>()
            .startWith(
              readQuery(
                request,
                optimistic: optimistic,
              ),
            )
            .takeUntil(dataChanged)
            .concatWith([
          DeferStream(
            () => closed
                ? Stream.empty()
                : watchQuery(request, optimistic: optimistic),
          )
```

If I understand correctly the intent was that return with the current value now and after every update. I replaced the code above with the following and it worked perfectly.

```dart
return dataChanged
            .map(
              (d) => readQuery(
                request,
                optimistic: optimistic,
              ),
            )
            .startWith(
              readQuery(
                request,
                optimistic: optimistic,
              ),
            );
```

@smkhalsa do you think both codes do the same thing?

janosroden:


smkhalsa (approved):
LGTM

PR #333, opened 2022-03-02
fix: remove analyzer from ferry_generator

aaahrens:
Current Problem:
Ferry generator isn't compatible with a lot of updated generator packages because of this old analyzer entry.

Example output: 
```
Because auto_route_generator 3.2.3 depends on analyzer >=2.7.0 <4.0.0 and ferry_generator >=0.5.0-dev.5 depends on analyzer ^1.7.2, auto_route_generator 3.2.3 is incompatible with ferry_generator >=0.5.0-dev.5.
So, because app depends on both ferry_generator ^0.5.0-dev.9 and auto_route_generator 3.2.3, version solving failed.
pub get failed (1; So, because app depends on both ferry_generator ^0.5.0-dev.9 and auto_route_generator 3.2.3, version solving failed.)

```

It looks like the analyzer entry is vestigial since `dart analyze` does not require to be added as a library.

Solution:
* Delete the entry
* Update existing deps gql_code_builder and gql_exec

Tested: 
* re-ran `flutter pub get` in ferry_generator

Inside ferry_test_graphql
* Made ferry_generator pull from current repo
* Ran `flutter pub run build_runner build --delete-conflicting-outputs`
* Ran `flutter test`

### DIFF
--- a/packages/ferry_exec/pubspec.yaml
+++ b/packages/ferry_exec/pubspec.yaml
@@ -7,7 +7,7 @@ environment:
   sdk: '>=2.12.0 <3.0.0'
 dependencies:
   gql: ^0.13.0
-  gql_exec: ^0.3.0
+  gql_exec: ^0.4.0
   gql_link: ^0.4.0
   meta: ^1.3.0
   async: ^2.5.0
--- a/packages/ferry_test_graphql/pubspec.yaml
+++ b/packages/ferry_test_graphql/pubspec.yaml
@@ -6,11 +6,11 @@ environment:
   sdk: '>=2.12.0 <3.0.0'
 dependencies:
   gql: ^0.13.0
-  gql_exec: ^0.3.0
+  gql_exec: ^0.4.0
   ferry_exec: ^0.1.5-dev.1
   built_value: ^8.0.4
   built_collection: ^5.0.0
-  gql_code_builder: ^0.3.0-alpha.0
+  gql_code_builder: ^0.5.0
 dev_dependencies:
   test: ^1.16.8
   build_runner: ^2.0.2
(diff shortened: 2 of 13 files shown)


aaahrens:
cc @smkhalsa I know this is blocking a few people

packages/ferry_exec/pubspec.yaml
```diff
@@ -17,6 +17,6 @@ dependencies:
 dev_dependencies:
   test: ^1.16.8
   pedantic: ^1.11.0
-  ferry_test_graphql: ^0.1.5-dev.2
+  ferry_test_graphql: 0.1.5-dev.3
```
smkhalsa, line 20: Please leave the ferry package versions. I'll bump them all together and publish new versions

packages/ferry_test_graphql/pubspec.yaml
```diff
@@ -6,14 +6,14 @@ environment:
...
   built_value: ^8.0.4
   built_collection: ^5.0.0
-  gql_code_builder: ^0.3.0-alpha.0
+  gql_code_builder: ^0.5.0
 dev_dependencies:
   test: ^1.16.8
   build_runner: ^2.0.2
   pedantic: ^1.11.0
-  ferry_generator: ^0.5.0-dev.9
+  ferry_generator: 0.5.0-dev.10
```
smkhalsa, line 18: Same thing here. Please leave the ferry package versions. I'll bump them all together and publish new versions

packages/ferry_exec/pubspec.yaml
aaahrens: Unbumped

packages/ferry_test_graphql/pubspec.yaml
aaahrens: Unbumped

smkhalsa (approved):

Issue #341, opened 2022-03-22
Request duplication issue


turtlepigg:
Making a request with the ferry client, repeats identical requests that were made in the past on that client. It doesn't seem to matter if it's a mutation or a query, running an identical request more than once where the request has the exact same data makes it start duplicating itself. In my case:
-I have a ferry client that's instantiated once in my class 
-I have a button that calls a function which tells that client to run my request and listen for the response
-First time I push the button it works as expected, I wait for the response, get a single response with the data I wanted
-Afterwards I push the button again which asks that same client to do the same thing
-The function is called once, but the request is executed twice for some reason and I get 2 response.
-If I push the button again, the function is called once, but the request is executed 3 times and I get 3 responses
-This goes on and on, always executing as many times as I've made that request in the past on the client

  ```
late Client client;

  @override
  void onInit() {
    super.onInit();
    var link = HttpLink(
      "https://rickandmortyapi.com/graphql",
    );
    client = Client(link: link);
  }
  
void getRickAndMortyCharacters(){
    print('Fetch');
    var charactersRequest = GCharactersReq(
          (b) => b
    );
    graphqlClient.request(charactersRequest).listen((event) {
      print('Got a fetch reply');
      if(event.hasErrors){
        event.printError();
        event.operationRequest.printError();
        print(event.linkException.toString());
        // print(event.printError());
      }
      else {
        print('It was successful');
        print(event.data.toString());
      }
    });
  } 
```

rae-yoder:
I am also seeing this issue when firing a mutation in my project, almost verbatim to what you are describing. Any luck with a workaround or solution?

turtlepigg:
@rae-yoder I did find a workaround, you just need to add a unique request ID each time you fire off the request so that they never look the exact same as each other. Kind of annoying to have to add the extra boiler plate to each request just to stop this duplicate request repetition behavior but it works for now. Hopefully this will get fix though. Here's an example where I use the https://pub.dev/packages/uuid package to generate a UUID for each request:
 var userRequest =  GUserReq(
          (b) => b
        ..vars.first = 10
        ..vars.offset = 0
        ..requestId = Uuid().v1(),

    );

warrenisarobot:
@smkhalsa I see you have done the most work on this file and wanted your feedback on the PR to address this issue.

warrenisarobot:
@turtlepigg a fix for this was merged in #346 , do you think we can close this now?

Tobiaswk:
[...]

PR #346, opened 2022-04-12
Updated RequestController duplciate request logic

warrenisarobot:
Updated logic in RequestControllerTypedLink to check the stream for the same instance, rather than equivalent, requests.  This was creating duplicate network requests if the a new request was created with the same parameters that had been issued previously.  This affects both queries and mutations.

This addresses the issue in #341


smkhalsa:
Thanks for the PR. Can you please add a test for this?

smkhalsa:
Thanks @warrenisarobot!

PR #359, opened 2022-07-22
fix: Streamline typename handling

budde377:
This PR streamlines how `dataId`'s are resolved throughout the normalization. Right now operation types are handled specially for operations, but not for fragments, or nested fields with the same type.

Assume we have the schema

```graphql
type Query {
  query: Query
  name: String
}
```

then normalizing the fragment

```graphql
fragment F on Query {
  __typename
  name
}
```
will throw an error unless a custom `dataIdFromObject` is specified. This while the above can be normalized just as easily as

```graphql
query Q {
  __typename
  name
}
```

Furthermore, this problem is also present with fields having the operation type as their type. Take the query

```graphql
Mutation Q2 {
  __typename
  q { name __typename }
}
```

normalizing this should yield

```json
{"query": {"__typename": "Query", "name": "Bob", "q": {"$ref": "Query"} }}
```

but it will yield

```json
{"query": {"__typename": "Query", "q": {"__typename": "Query", "name": "Bob"} }}
```

I.e., we're missing out on updating the cache correctly. Returning `Query` from e.g. a mutation can be a nifty way to allow advanced cache updates.

Finally, this PR fixes a bug where nested objects within themselves will not be normalized correctly. 

**NOTICE, THIS CHANGE IS BREAKING**: All users providing `dataIdFromObject` will have to handle operation types appropriately.

### DIFF
--- a/packages/normalize/lib/src/denormalize_fragment.dart
+++ b/packages/normalize/lib/src/denormalize_fragment.dart
@@ -2,6 +2,7 @@ import 'package:gql/ast.dart';
 import 'package:normalize/normalize.dart';
 
 import 'package:normalize/src/config/normalization_config.dart';
+import 'package:normalize/src/utils/constants.dart';
 import 'package:normalize/src/utils/get_fragment_map.dart';
 import 'package:normalize/src/utils/resolve_data_id.dart';
 import 'package:normalize/src/utils/add_typename_visitor.dart';
@@ -33,7 +34,7 @@ Map<String, dynamic>? denormalizeFragment({
   bool addTypename = false,
   bool returnPartialData = false,
   bool handleException = true,
-  String referenceKey = '\$ref',
+  String referenceKey = kDefaultReferenceKey,
   Map<String, Set<String>> possibleTypes = const {},
 }) {
   if (addTypename) {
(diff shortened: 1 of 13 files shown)


budde377:
@smkhalsa, I know you're looking for maintainers. `graphql_flutter` is still depending on this project, do you think it makes sense for us to fork it and drive our own version?

knaeckeKami:
Hi! thank you for your PR. I am a new maintainer of ferry. I would like to keep resources pooled and avoid people using several forks of the same package when it's not needed.

Unfortunately, I am not yet up to speed with the internals of cache normalization. 

Since you mentioned that this is a breaking change:
Do you know what would happen to ferry-cache if we merged this? 

Would it still work? Would users trying to read the old cache data after this PR still be able to read their data?
If not, is it feasible to migrate to data?

budde377:
Good to meet you @knaeckeKami! As I mentioned in the description this will change behaviour for all users providing their own `dataIdFromObject` as they'll have to explicitly generate a data id for the operation type (`Query`, `Mutation`, etc.). Whether this'll impact ferry I don't know, I'm not using ferry. 

Can you provide me with an eta on when you'll be up to speed and able to indicate if this is an acceptable change? This is impacting users on graphql_flutter of which I'm a maintainer.

knaeckeKami:
I see that you put a lot of thought into this and I will make sure to review it appropriately.
However, I first need to get into a few other issues and thoroughly investigate the implications on ferry_cache, especially regarding the persistent cache. 

I understand that, since you depend on this normalize, you would like to have it merged and released to pub.dev as soon as possible.

I propose the following:

- for users needing the fix immediately, tell them to add a dependency_override to your fork 
- I will fix the CI (see errors on the checks) then merge it and then release an alpha version of `normalize`  that you can depend on. eta: 7-14 days. This should unblock you. 
- Once I am aware of the impact on ferry_cache/ferry_hive_store users (with persistent cache), I can tell you more regarding a non-alpha release. I can not commit to a timeframe at this point though.

Does this sound acceptable to you?

knaeckeKami:
published normalize: ^0.7.0-dev.1

PR #368, opened 2022-08-03
fix: Preserve backwards compatibility

budde377:
This change will ensure backward compatibility by defaulting the typename to the `rootTypeName` when we can't resolve the data ID for a root object. 

After some consideration, I believe this is a better implementation. We HAVE to write the root objects otherwise no nested objects will be saved.

This, I believe, also means that we can release this as a minor change.


knaeckeKami:
LGTM

knaeckeKami:
published normalize: 0.6.1

PR #385, opened 2022-08-22
fix(cache): Fix memory leak.

### LINKED ISSUE #343: The cache causes memory leak

GP4cK:
The issue was caused by this line: https://github.com/gql-dart/ferry/blob/201980f313a98aa94ef45146ff229f24a95c3125/packages/ferry_cache/lib/src/cache.dart#L105

As per the [docs](https://api.flutter.dev/flutter/dart-async/Stream/isEmpty.html), isEmpty `waits for the first element of this stream`. The problem is that `operationDataChangeStream` and `fragmentDataChangeStream` end with `.skip(1)` so the stream never emits and the Future is pending forever.

My fix seems extremely simple, so maybe I missed some complexity but all the tests are passing...
Also, I don't know how to write an additional test to make sure that we don't leak.

If you want to manual test, I've created a branch where I added the original memory-leak repo: https://github.com/GP4cK/ferry/tree/memory-leak-example (it overrides the ferry dependencies to make sure to use my version of the code).

Closes #343

### DIFF
--- a/packages/ferry_cache/lib/src/cache.dart
+++ b/packages/ferry_cache/lib/src/cache.dart
@@ -1,14 +1,14 @@
 import 'dart:async';
+
+import 'package:ferry_exec/ferry_exec.dart';
+import 'package:ferry_store/ferry_store.dart';
 import 'package:meta/meta.dart';
-import 'package:pedantic/pedantic.dart';
-import 'package:rxdart/rxdart.dart';
 import 'package:normalize/normalize.dart';
 import 'package:normalize/utils.dart' as utils;
-import 'package:ferry_store/ferry_store.dart';
-import 'package:ferry_exec/ferry_exec.dart';
+import 'package:rxdart/rxdart.dart';
 
-import './operation_data_change_stream.dart';
 import './fragment_data_change_stream.dart';
+import './operation_data_change_stream.dart';
 
 class Cache {
   final Map<String, TypePolicy> typePolicies;
@@ -95,33 +95,12 @@ class Cache {
     required Stream Function() getChangeStream,
     required TData? Function() getData,
   }) {
-    late StreamController<TData?> sc;
-
-    Future<void> addNext() async {
-      try {
-        if (sc.isClosed) return;
-
-        sc.add(getData());
-        if (await getChangeStream().isEmpty) {
-          unawaited(sc.close());
-          return;
-        }
-
-        unawaited(addNext());
-      } catch (err, st) {
-        if (!sc.isClosed) {
-          sc.addError(err, st);
-          unawaited(addNext());
-        }
-      }
-    }
-
-    sc = StreamController<TData?>(
-      onListen: () => addNext(),
-      onCancel: () => unawaited(sc.close()),
-    );
-
-    return sc.stream;
+    return getChangeStream()
+    // We add null at the beginning of the stream to trigger the initial getData().
+    // getChangeStream = operationDataChangeStream or fragmentDataChangeStream and
+    // they both end with .skip(1).
+    .startWith(null) 
+    .map((_) => getData());
   }
 
   /// Reads denormalized data from the Cache for the given operation.
--- a/packages/ferry_cache/lib/src/cache.dart
+++ b/packages/ferry_cache/lib/src/cache.dart
@@ -96,11 +96,11 @@ class Cache {
     required TData? Function() getData,
   }) {
     return getChangeStream()
-    // We add null at the beginning of the stream to trigger the initial getData().
-    // getChangeStream = operationDataChangeStream or fragmentDataChangeStream and
-    // they both end with .skip(1).
-    .startWith(null) 
-    .map((_) => getData());
+        // We add null at the beginning of the stream to trigger the initial getData().
+        // getChangeStream = operationDataChangeStream or fragmentDataChangeStream and
+        // they both end with .skip(1).
+        .startWith(null)
+        .map((_) => getData());
   }
 
   /// Reads denormalized data from the Cache for the given operation.


knaeckeKami:
Thanks!  
Yes, this is indeed a leak. The StreamController is never closed when the listener already stopped listening but no changes are ever emitted.

I am not quite sure of the proposed fix though. I would like to suggest something less risky to introduce unwanted behavior changes:
instead of
```dart
        if (await getChangeStream().isEmpty) { 
```
let's additionally wait for the `done`event of the StreamController:

```dart
        final isEmptyFuture =  getChangeStream().isEmpty;
        final streamControlleDoneFuture = sc.done.then((value) => true);

        if (await Future.any([isEmptyFuture, streamControlleDoneFuture]) ) {
          unawaited(sc.close());
          return;
        }
```

What do you think about this?

This will finish the closure if the client stops listening to the stream controller, even if there are no further changes to a relevant section to the cache.

----

Also, I actually see now another leak here:

`addNext()` is a potentially unbounded recursive function, on every change to the cache, the recursion goes one level deeper, which let's the call stack grow, another form of a memory leak. 

This is not reproducible with the example as the cache does not actually change after the first request. 

I think this should be fixable by rewriting `addNext()` as loop instead of recursive function.

I'll leave it to you if you also want to tackle it or not, I would also accept the MR with just a solution to the `await getChangeStream()` call that never completes and fix the other issue later.

GP4cK:
Thanks for the review.

> the changestream actually emits an event in other cases, as you can see in the test cases of the cache

Do you have a specific example? I looked at `packages/ferry_cache/test/test/data_change_streams_test.dart` and there are a few tests that make sure the stream doesn't emit when it's not necessary (unused operation fields, unrelated data etc.).
I could also add a `.distinct()` before or after the `.map((_) => getData())`...

GP4cK:
Hey @knaeckeKami sorry if I seem pushy but when do you think this could be merged?
We've got a couple of reports in Sentry saying that our app crashed because of OOM and a few users have reported that the web freezes after using it for a while and I wonder if it's because of this issue...
Cheers

knaeckeKami:
Hi!

I plan to have another proper review this weekend. I found a potential issue/change in behavior and need some time to check if it is relevant or not. It is about getChangeStream(), since it is now only called once I am not sure if it will behave correctly if the dataIds change after the stream was started. But I'm not sure if this is a problem yet.

But you can always use your version via dependency_overrides:

```yaml
ferry_cache:
   git:
      url: https://github.com/GP4cK/ferry.git
      ref: fix-memory-leak
      path: packages/ferry_cache
```

GP4cK:
Cool thanks for the update!

knaeckeKami:
I created a bunch of additional tests and ran them against the current and the new implementation.
Good news: the only difference in behavior that I found is actually another bug in the current implementation ;)

So, LGTM

knaeckeKami:
published pre ferry_cache: 0.6.2-dev.0+1

GP4cK:
> Good news: the only difference in behavior that I found is actually another bug in the current implementation ;)

What's the other bug? Maybe I can help looking into it too :)

knaeckeKami:
Sorry, I was not clear: I added additional test, and the test uncovered another bug in the previous implementation with the memory leak. Your MR fixed this bug also. 

It was this test: https://github.com/gql-dart/ferry/blob/master/packages/ferry_cache/test/watch_test.dart#L112, the cache stream would not recover from an error previously, now it does.

GP4cK:
Ah ok. Cool :)

knaeckeKami:
Hi! 

This PR added a regression.
A made a mistake in the test that was supposed to check for this.

This test (in watch_test in ferry_cache) fails now, but succeeds with the previous implementation:

```dart
 test('can receive updates when child objects are updated by other queries',
            () async {
          final cache = Cache();
          cache.writeQuery(reviewsReq, reviewsData);

          final updatedReview =
          reviewsData2.reviews![1].rebuild((b) => b.commentary = 'first');

          expect(
              cache.watchQuery(reviewsReq),
              emitsInOrder([
                reviewsData,
                reviewsData2,
                reviewsData2.rebuild((b) => b..reviews[1] = updatedReview),
                emitsDone,
              ]));

          await Future.delayed(Duration.zero);

          cache.writeQuery(reviewsReq, reviewsData2);

          await Future.delayed(Duration.zero);

          cache.writeQuery(
              GReviewsByIDReq((b) => b..vars.id = reviewsData2.reviews![1].id),
              GReviewsByIDData((b) => b.review
                ..id = updatedReview.id
                ..stars = updatedReview.stars
                ..createdAt = updatedReview.createdAt
                ..seenOn = updatedReview.seenOn.toBuilder()
                ..commentary = updatedReview.commentary));

          await Future.delayed(Duration.zero);

          await cache.dispose();
        });
```

The regression is: The cache does not emit updates, when objects, that where only added after the initial query was started, change. If more objects are added the cache will emit them, but it will not emit them when their data updates.

The reason for this is that since now `operationDataChangeStream`/`fragmentDataChangeStream` is only called once, they will keep their list of data ids and not watch newly created data objects when they are added.

GP4cK:
Oh ok thanks. Sorry I missed this 😞

knaeckeKami:
No worries, I missed it too ;)

PR #401, opened 2022-09-04
fix(ferry_cache): fix regression in watch() methods 

knaeckeKami:
this fixes the regression I was talking about in #385 in https://github.com/gql-dart/ferry/pull/385#issuecomment-1229184574. It also fixes the test that was not testing the thing it was supposed to test.

@GP4cK If you have suggestions for better implementations, please share them

### DIFF
--- a/packages/ferry_cache/lib/src/cache.dart
+++ b/packages/ferry_cache/lib/src/cache.dart
@@ -95,12 +95,41 @@ class Cache {
     required Stream Function() getChangeStream,
     required TData? Function() getData,
   }) {
-    return getChangeStream()
-        // We add null at the beginning of the stream to trigger the initial getData().
-        // getChangeStream = operationDataChangeStream or fragmentDataChangeStream and
-        // they both end with .skip(1).
-        .startWith(null)
-        .map((_) => getData());
+    late StreamController<TData?> sc;
+
+    Future<void> addNext() async {
+      while (true) {
+        try {
+          if (sc.isClosed) return;
+
+          sc.add(getData());
+          final isEmptyFuture = getChangeStream().isEmpty;
+          final streamControllerDoneFuture = sc.done.then((value) => true);
+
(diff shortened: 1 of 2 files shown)


packages/ferry_cache/lib/src/cache.dart
```diff
@@ -95,12 +95,41 @@ class Cache {
...
+            return;
+          }
+        } catch (err, st) {
+          if (!sc.isClosed) {
+            sc.addError(err, st);
+          }
+          //continuing immediately likely would result in an endless loop where
+          //we emit the same error over and over again
+          //so we wait for the next change
+          final isEmptyFuture = getChangeStream().first;
```
GP4cK, line 120: Here should it be `.isEmpty` or `.first`?

GP4cK:
I tried merging your branch into the one where I added the memory leak example and it seems to leak though :/

Here's the branch: https://github.com/GP4cK/ferry/tree/memory-leak-example-2

<details>
<summary>Screenshot</summary>

image

</details>

packages/ferry_cache/lib/src/cache.dart
knaeckeKami: probably .first, good call

knaeckeKami:
Ah yeah true, of course this leaks as well. We need to cancel the subscription to the changeStream if the stream controller closes. .isEmpty will still listen to the stream. this should be fixed fairly  easily.

knaeckeKami:
Now the code does not leak and supports changing data ids. 
But it's very ugly. Need to sleep over this to figure out how to clean this up.

GP4cK:
> Now the code does not leak and supports changing data ids. But it's very ugly. Need to sleep over this to figure out how to clean this up.

I've added a draft PR based on your `todo`. PTAL: https://github.com/gql-dart/ferry/pull/403

knaeckeKami:
Closing in favor of #403

PR #403, opened 2022-09-10
fix(ferry_cache): fix regression in watch() methods by making changeStream emit when new item is added to the cache

GP4cK:
@knaeckeKami this is an alternative to https://github.com/gql-dart/ferry/pull/401/files to fix the regression in the watch method based on your todo:

```dart
// todo: refactor operationDataChangeStream / fragmentChangeDataStream so that they can
// deal with added ids, so we don't have to maintain this ugly block below
```

I've only modified the `operationDataChangeStream` for now to test the idea with you. I would need to update the `fragmentDataChangeStream` too before this gets merged.

~~The issue that I have though is that in the tests, [we are expecting the watchQuery() to emit a done event](https://github.com/gql-dart/ferry/blob/0c3cccc787d85135f2189a670d485614a5cf51fa/packages/ferry_cache/test/watch_test.dart#L163) but this implementation doesn't. If we comment the line, the test will pass. From what I understand, the done event should be emitted when the cache is disposed, but currently we have no way to subscribe to this event. I don't even understand how that works on master.~~
**Edit**: nvm, I figured it out by looking at the implementation of `CombineLatestStream`. See this commit [852046d](https://github.com/gql-dart/ferry/pull/403/commits/852046dd864a2c57825c12b3d0e9c154bac73401).

The first commit is just copying your failing test.
I suggest to review the second commit by ignoring the whitespaces, it'll make it easier to see the real changes.

### DIFF
--- a/packages/ferry_cache/lib/src/operation_data_change_stream.dart
+++ b/packages/ferry_cache/lib/src/operation_data_change_stream.dart
@@ -102,7 +102,9 @@ Stream<Set<String>> operationDataChangeStream<TData, TVars>(
             changed.clear();
             return result;
           },
-        )
+        ).doOnDone(() {
+          dataIdStreamController.close();
+        })
 
             /// Skip the first result since this returns the existing data
             .skip(1);
--- a/packages/ferry_cache/lib/src/operation_data_change_stream.dart
+++ b/packages/ferry_cache/lib/src/operation_data_change_stream.dart
@@ -55,7 +55,7 @@ Stream<Set<String>> operationDataChangeStream<TData, TVars>(
 
         return dataIds;
       })
-      .distinct()
+      .distinct((prev, next) => const SetEquality<String>().equals(prev, next))
       .switchMap((dataIds) {
         /// IDs that have changed
         final changed = <String>{};
--- a/packages/ferry_cache/lib/src/operation_data_change_stream.dart
+++ b/packages/ferry_cache/lib/src/operation_data_change_stream.dart
@@ -1,11 +1,12 @@
-import 'package:normalize/policies.dart';
-import 'package:normalize/utils.dart';
-import 'package:normalize/normalize.dart';
+import 'dart:async';
+
 import 'package:collection/collection.dart';
+import 'package:ferry_exec/ferry_exec.dart';
 import 'package:ferry_store/ferry_store.dart';
+import 'package:normalize/normalize.dart';
+import 'package:normalize/utils.dart';
 import 'package:rxdart/rxdart.dart';
 
-import 'package:ferry_exec/ferry_exec.dart';
 import './utils/data_for_id_stream.dart';
 import './utils/operation_root_data.dart';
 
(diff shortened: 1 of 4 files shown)


packages/ferry_cache/lib/src/operation_data_change_stream.dart
```diff
@@ -31,67 +32,81 @@ Stream<Set<String>> operationDataChangeStream<TData, TVars>(
     typePolicies,
   );
 
-  final dataIds = <String>{};
+  final dataIdStreamController = StreamController<void>();
```
GP4cK, line 35: The issue of the regression is that `dataIds` was a map that was only populated once by the call to `denormalizeOperation`. 
Later, if new items are added to the cache, their dataId wouldn't be added to the map and so they wouldn't be watched.

So the fix is to make `dataIds` the result of a stream so we can watch new items.

packages/ferry_cache/lib/src/operation_data_change_stream.dart
```diff
@@ -31,67 +32,81 @@ Stream<Set<String>> operationDataChangeStream<TData, TVars>(
...
+          returnPartialData: true,
+          dataIdFromObject: dataIdFromObject,
+          possibleTypes: possibleTypes,
+        );
 
-  /// IDs that have changed
-  final changed = <String>{};
+        return dataIds;
+      })
+      .distinct()
```
GP4cK, line 58: Here we use `.distinct()` because if the dataIds map hasn't changed, there's no reason to emit a new event.

packages/ferry_cache/lib/src/operation_data_change_stream.dart
```diff
@@ -31,67 +32,81 @@ Stream<Set<String>> operationDataChangeStream<TData, TVars>(
...
-        .doOnData((_) => changed.add(dataId));
-  });
+          return stream.distinct(
+            (prev, next) {
+              final areEqual = const MapEquality().equals(prev, next);
+              if (!areEqual) {
+                // Maybe a new element was added to an array,
+                // we need to recompute the dataIds
+                dataIdStreamController.add(null);
+              }
```
GP4cK, line 92: We now that `prev` and `next` are maps so we can use `MapEquality` instead of `DeepCollectionEquality`.

If the result are different, we may need to recompute the `dataIds` map, so we emit a new event with the dataIdStreamController. If the maps are the same, it will be ignored thanks to the `.distinct()`.

packages/ferry_cache/lib/src/operation_data_change_stream.dart
```diff
@@ -31,67 +32,81 @@ Stream<Set<String>> operationDataChangeStream<TData, TVars>(
...
+            return result;
+          },
+        )
 
-      /// Skip the first result since this returns the existsing data
-      .skip(1);
+            /// Skip the first result since this returns the existing data
+            .skip(1);
+      });
+  dataIdStreamController.add(null);
```
GP4cK, line 110: Add an initial event to the dataIdStream to trigger the computation of the `dataIds` map.

packages/ferry_cache/lib/src/operation_data_change_stream.dart
knaeckeKami: `distinct()` does not do anything here as we create a new dataId Set every time though and Sets don't implement `==`. To really filter updates, we would have to pass the equals method to use `SetEquality()` or something similar.

knaeckeKami:
Aside from the one comment, looks good to me! I 

I would be great if you also applied the changes also to `fragmentDataChangeStream`.

packages/ferry_cache/lib/src/operation_data_change_stream.dart
GP4cK: Good catch. Fixed it in [37c4e20](https://github.com/gql-dart/ferry/pull/403/commits/37c4e204b2dd8a9a040e7bd890133b8f2b95319a)

GP4cK:
> It would be great if you also applied the changes also to `fragmentDataChangeStream`.

I've added a failing test on the fragmentDataChangeStream in [874a3b5](https://github.com/gql-dart/ferry/pull/403/commits/874a3b5ea6637efaed1500435f4890adff72edf6) and I updated the code in [38a3b80](https://github.com/gql-dart/ferry/pull/403/commits/38a3b8070f280371724b256b596031ebde3d27b6).

Thanks for your review!

Issue #404, opened 2022-09-12
Query in tests - pending timer error


ValentinVignal:
The current implementation of `RequestControllerTypedLink` makes the tests fail when a query is made because there are pending timers.

Here is what I have:

(you can also checkout the code at https://github.com/ValentinVignal/flutter_app_stable/tree/ferry/timer-in-tests)

```graphql
# schema.graphql
type Pokemon {
  id: ID!
  name: String!
}

type Query {
  pokemons: [Pokemon!]!
}
```

```graphql
# pokemon_query.graphql
query Pokemons {
  pokemons {
    id
    name
  }
}
```

In a test file, I create a stream builder that listens to the stream from the request `Pokemons`:

```dart
final _client = Client(link: HttpLink('http'));

class _MyWidget extends StatefulWidget {
  const _MyWidget({Key? key}) : super(key: key);

  @override
  State<_MyWidget> createState() => __MyWidgetState();
}

class __MyWidgetState extends State<_MyWidget> {
  late final _stream = _client.request(GPokemonsReq());

  @override
  Widget build(BuildContext context) {
    return StreamBuilder(
      stream: _stream,
      builder: (context, snapshot) {
        return const SizedBox();
      },
    );
  }
}

void main() {
  testWidgets('Counter increments smoke test', (WidgetTester tester) async {
    await tester.pumpWidget(const _MyWidget());
  });
}
```

When run, I get the error:

```zsh
00:03 +0: Pending timers test                                                                                                                                                                                                         
Pending timers:
Timer (duration: 0:00:00.000000, periodic: false), created:
#0      new FakeTimer._ (package:fake_async/fake_async.dart:308:62)
#1      FakeAsync._createTimer (package:fake_async/fake_async.dart:252:27)
#2      FakeAsync.run.<anonymous closure> (package:fake_async/fake_async.dart:185:19)
#6      RequestControllerTypedLink.request.<anonymous closure> (package:ferry/src/request_controller_typed_link.dart:46:16)
#7      _DoStreamSink.onData (package:rxdart/src/transformers/do.dart:31:16)
#29     _WhereTypeStreamSink.add (package:rxdart/src/transformers/where_type.dart:11:19)
#47     FakeAsync.flushMicrotasks (package:fake_async/fake_async.dart:197:32)
#48     AutomatedTestWidgetsFlutterBinding.pump.<anonymous closure> (package:flutter_test/src/binding.dart:1059:26)
[...]

PR #406, opened 2022-09-15
Fix/pending timer

GP4cK:
Fixes https://github.com/gql-dart/ferry/issues/404

The first commit is to add the test which is failing in the issue.
The second one refactors the request method to not use a Future.delayed (probably easier to review when ignoring the whitespaces)


knaeckeKami:
this is a duplicate of #407 right?

GP4cK:
> this is a duplicate of #407 right?

Yes. I'll just close it.

PR #407, opened 2022-09-15
fix(ferry): Fix pending timer in test

GP4cK:
Fixes https://github.com/gql-dart/ferry/issues/404

The first commit is to add the test which is failing in the issue.
The second one refactors the request method to not use a Future.delayed (probably easier to review when ignoring the whitespaces)


GP4cK:
Mmm I tried reopening the same PR with lower-cased commits but the github action is still failing. Any idea why? I'll try to investigate.

knaeckeKami:
The PR title also needs to match the conventional commit style because that's what's used as commit message when a commit gets merged using the "Squash and merge" strategy. I took the liberty of renaming the title.

knaeckeKami:
Looks good to me, but as this touches the `updateResult` functionality, which I never used and which is not well tested, 
I'll add some more tests to this

knaeckeKami:
This test passes with the current implementation but fails with the updated one:

```dart
group('update result with real client', () {

    test("real client test", () async
    {
      final client = Client(link: _AutoResponderTerminalLink());

      final req1 = req.rebuild((b) => b..requestId = '123');

      final stream = client.request(req1);

      final reviews1 = dataForRequest(req1).reviews!;


      final req2 = req1.rebuild((b) => b..vars.offset = 3);

      final reviews2 = reviews1.rebuild(
            (b) => b..addAll(dataForRequest(req2).reviews!),
      );

      expect(stream, emitsInOrder(
          [
            isA<OperationResponse<GReviewsData, GReviewsVars>>()
                .having((p) => p.data!.reviews, 'reviews', reviews1),
            isA<OperationResponse<GReviewsData, GReviewsVars>>()
                .having((p) => p.data!.reviews, 'reviews', reviews2),
            emitsDone
          ]
      ));

      await Future.delayed(Duration.zero);


      client.requestController.add(req2);

      await Future.delayed(Duration.zero);

      await client.dispose();
    });

```

with

```dart

class _AutoResponderTerminalLink extends Link {
  @override
  Stream<Response> request(
    Request req, [
    forward,
  ]) =>
      Stream.value(Response(
          response: {},
          data: GReviewsData((b) => b
            ..reviews.addAll([
              for (int i = req.variables["offset"] as int;
                  i < ((req.variables["offset"] as int) + req.variables["first"] as int);
                  i++)
                GReviewsData_reviews(
                  (b) => b
                    ..id = '$i'
                    ..stars = 5,
                )
            ])).toJson()));
}
```

in `request_controller_typed_link_test.dart`

It seems it tries to subscribe to the same stream twice, even though it's a single subscription stream.

```
dart:async                                                 _StreamImpl.listen
package:rxdart/src/streams/combine_latest.dart 315:25      CombineLatestStream._buildController.<fn>.<fn>
dart:_internal                                             ListIterable.toList
package:rxdart/src/streams/combine_latest.dart 335:12      CombineLatestStream._buildController.<fn>
dart:async                                                 StreamView.listen
package:rxdart/src/streams/connectable_stream.dart 123:13  ValueConnectableStream._connection
package:rxdart/src/streams/connectable_stream.dart         ValueConnectableStream._connection
package:rxdart/src/streams/connectable_stream.dart 154:46  ValueConnectableStream.refCount.<fn>
dart:async                                                 _StreamImpl.listen
package:rxdart/src/streams/defer.dart 50:19                DeferStream.listen
dart:async                                                 StreamView.listen
package:rxdart/src/transformers/switch_map.dart 25:40      _SwitchMapStreamSink.onData
dart:async                                                 _BroadcastStreamController.add
package:rxdart/src/transformers/do.dart 40:10              _DoStreamSink.onData
dart:async                                                 _EventSinkWrapper.add
package:rxdart/src/transformers/where_type.dart 11:19      _WhereTypeStreamSink.add

Bad state: Stream has already been listened to.
```

GP4cK:
Hello @knaeckeKami, thanks for the additional test!
`current` can be subscribed to twice because it is first returned as a result of the `switchMap` and then, if another request comes in with a non-null `updateResult`, then `current` becomes `previous` and it will be listened again via the `CombineLatestStream`. This is why we have to use `.shareValue()` (in today's implementation too).

Your test passes if I make all the `current` stream ValueStream (see https://github.com/gql-dart/ferry/pull/407/commits/e0eed734cca303c0e4c4ea63db5a0719b846f03e). However, a couple of other tests then fail because of timeouts. I think somehow, one of the value stream never gets closed.

I managed to find another way by refactoring the code to use a StreamController so I can control when the streams should be closed (https://github.com/gql-dart/ferry/pull/407/commits/b9705f0c399fa6f8066f96fb3f704106b0662cb8). However, another test would break (`cancelling subscription and doing new request resets state`) because we were reaching the `expect(lastResponse, reviews3);` before the stream had time to emit/combine all the reviews. I'm pretty sure that it's because of the `.pipe()` which adds the events asynchronously. 
To make all of them pass, I have to use a [SynchronousStreamController](https://api.dart.dev/dev/2.8.0-dev.7.0/dart-async/SynchronousStreamController-class.html) with this line:

```dart
prev = StreamController(sync: true); // you may remove sync: true and run the tests to check
```

I've never used it before and the dart doc says to use it with caution but it seems to work...
Let me know what you think.
Cheers

knaeckeKami:
Would the failing test succeed if you add another Future.wait(Duration.zero) (or two?) before the failing expect?

I am okay with a slight increase of latency of another microtask-loop, because awaiting the actual network response would dominate anyway.

knaeckeKami:
Also, can we maybe simplify this using the scan operator? https://pub.dev/documentation/rxdart/latest/rx/ScanExtension/scan.html

GP4cK:
> Would the failing test succeed if you add another Future.wait(Duration.zero) (or two?) before the failing expect?

No the test still fails. It passes however if I switch the 2 statements:
```dart
// KO
await sub.cancel();
await Future.delayed(Duration.zero);
// OK
await Future.delayed(Duration.zero);
await sub.cancel();
```

I'll try with the scan operator later this week-end.

GP4cK:
Trying with `.scan()`, I came up with this implementation:

<details>
<summary>Implementation</summary>

```dart
@override
  Stream<OperationResponse<TData, TVars>> request<TData, TVars>(
    request, [
    forward,
  ]) {
    var initial = true;

    return requestController.stream
        .whereType<OperationRequest<TData, TVars>>()
        .where(
          (req) => req.requestId == null
              ? identical(req, request)
              : req.requestId == request.requestId,
        )
        .scan<Stream<OperationResponse<TData, TVars>>>(
          (acc, next, index) {
            if (next.updateResult == null || index == 0) {
              return forward!(next).shareValue();
            }

            return CombineLatestStream.combine2<
                OperationResponse<TData, TVars>,
                OperationResponse<TData, TVars>,
                OperationResponse<TData, TVars>>(
              acc,
              forward!(next),
              (previous, response) => OperationResponse<TData, TVars>(
                operationRequest: response.operationRequest,
                data: response.operationRequest.updateResult!(
                  previous.data,
                  response.data,
                ),
                dataSource: response.dataSource,
                linkException: response.linkException,
                graphqlErrors: response.graphqlErrors,
              ),
            ).shareValue();
          },
          Stream.value(
            OperationResponse<TData, TVars>(operationRequest: request),
          ),
        )
        .switchMap((stream) => stream)
        .doOnListen(
          () {
            if (initial && request.executeOnListen) {
              scheduleMicrotask(() => requestController.add(request));
            }
            initial = false;
          },
        );
  }
```
</details>

But then, this test fails `continues to track changes to previous` because of a timeout exception after 30s. I'm pretty sure it's because of the `.shareValue()` but I can't find a way to fix it... :/

knaeckeKami:
Hm.
What about replacing 

```dart
        final sub = prev?.listen(null);
        Future.delayed(Duration.zero).then((_) => sub?.cancel());
```
with
```dart
        final sub = prev?.listen(null);
        scheduleMicrotask(() => sub?.cancel());
```

This should keep the original behavior, and won't cause widget tests to fail because micro tasks do not have a pending timer.

GP4cK:
@knaeckeKami replacing `Future.delayed` with `scheduleMicrotask` indeed works! I wish I didn't spend so much time on this haha.
Well done sir.

knaeckeKami:
I feel you. 
It took me also a lot of time to understand to problem before I thought of that

Issue #426, opened 2022-09-26
Query in tests - pending timer error when policy is CacheAndNetwork


GP4cK:
Similar to https://github.com/gql-dart/ferry/issues/404 the issue was addressed for CacheFirst but not for CacheAndNetwork. I'll open a PR to fix it.

PR #427, opened 2022-09-26
refactor: FetchPolicyTypedLink.request() to avoid pending timers in tests

GP4cK:
Closes https://github.com/gql-dart/ferry/issues/426

### DIFF
--- a/packages/ferry/lib/src/fetch_policy_typed_link.dart
+++ b/packages/ferry/lib/src/fetch_policy_typed_link.dart
@@ -1,9 +1,7 @@
 import 'dart:async';
-import 'package:gql_link/gql_link.dart';
 import 'package:rxdart/rxdart.dart';
 import 'package:normalize/utils.dart';
 import 'package:gql/ast.dart';
-import 'package:ferry_cache/ferry_cache.dart';
 import 'package:ferry_exec/ferry_exec.dart';
 
 import 'package:ferry/src/optimistic_typed_link.dart';
@@ -103,7 +101,7 @@ class FetchPolicyTypedLink extends TypedLink {
               /// Temporarily add a listener so that [sharedNetworkStream] doesn't shut down when
               /// switchMap is updating the stream.
               final sub = sharedNetworkStream.listen(null);
-              Future.delayed(Duration.zero).then((_) => sub.cancel());
+              scheduleMicrotask(sub.cancel);
             },
           ),
         ).concatWith([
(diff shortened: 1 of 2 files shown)


packages/ferry/lib/src/fetch_policy_typed_link.dart
```diff
@@ -1,9 +1,7 @@
 import 'dart:async';
-import 'package:gql_link/gql_link.dart';
```
GP4cK, line 2: VSCode was warning me that "The import of 'package:gql_link/gql_link.dart' is unnecessary because all of the used elements are also provided by the import of 'package:ferry/src/gql_typed_link.dart'."

Same situation for ferry_cache below.

packages/ferry/lib/src/fetch_policy_typed_link.dart
```diff
@@ -94,28 +92,15 @@ class FetchPolicyTypedLink extends TypedLink {
...
           sharedNetworkStream
               .doOnData(_removeOptimisticPatch)
               .doOnData(_writeToCache)
-              .switchMap(
-                (networkResponse) => ConcatStream([
-                  Stream.value(networkResponse),
-                  _cacheTypedLink.request(operationRequest).skip(1),
-                ]),
-              )
+              .concatWith([_cacheTypedLink.request(operationRequest).skip(1)])
```
GP4cK, line 103: I tried using scheduleMicrotask like in the other PR instead of the Future.delayed(0) in the previous commit but it broke the test `can return multiple responses from link`.

Instead, I refactored the code to return a MergeStream where:
- The first stream is the cache stream, which will stop emitting as soon as the network stream emits
- The second stream emits the response and when it's done, it's concatenated with the cacheTypeLink (like in the previous implementation)

packages/ferry/lib/src/fetch_policy_typed_link.dart
```diff
@@ -1,18 +1,16 @@
...
 
-import 'package:ferry/src/optimistic_typed_link.dart';
-import 'package:ferry/src/gql_typed_link.dart';
 import 'package:ferry/src/cache_typed_link.dart';
+import 'package:ferry/src/gql_typed_link.dart';
+import 'package:ferry/src/optimistic_typed_link.dart';
+import 'package:ferry_exec/ferry_exec.dart';
+import 'package:gql/ast.dart';
+import 'package:normalize/utils.dart';
+import 'package:rxdart/rxdart.dart';
```
GP4cK, line 9: I have a setting in VSCode that sorts the imports alphabetically and forgot to turn it off. Let me know if you want me to revert the order.

GP4cK:
@knaeckeKami just pinging you. You think you could take a look at this?
I know most of my changes introduced regressions in the past but I'm pretty confident about that one 😅 
I hope all is well on your side.

knaeckeKami:
Yeah, took a bit of a break. Will look into this weekend. 

No worries, all is good ;)

knaeckeKami:
LGTM

PR #433, opened 2022-10-09
feat(normalize)!: respect skip/include directives when checking for partial data

### LINKED ISSUE #432: Support for @include and @skip directives in normalize
Is your feature request related to a problem? Please describe.
Attempting to use the @include or @Skip directives in `graphql_flutter` results in a PartialDataException. This would be a great feature for re-using queries in multiple places where some fields may not be needed in all cases.

Describe the solution you'd like
I'd like the @include directive to work as one would expect:
- If false, the decorated field or fragment in an operation is not resolved by the GraphQL server
and the @Skip directive to work as one would expect:
- If true, the decorated field or fragment in an operation is not resolved by the GraphQL server.

Describe alternatives you've considered
An alternative might be to create separate queries for each use case, but that defeats the goal of reusability.

Additional context
Thanks for considering!

knaeckeKami:
When checking for partial data, do not treat missing fields as partial data, if the field has a `skip` directive with an `if:` argument that evaluates to `true` or an `include` directive with an `if:` argument that evaluates to `false`
or if the field is in a fragment or an inline fragment with such a directive

fixes #432

### DIFF
--- a/packages/normalize/lib/src/utils/well_known_directives.dart
+++ b/packages/normalize/lib/src/utils/well_known_directives.dart
@@ -1,0 +1,64 @@
+import 'package:collection/collection.dart';
+import 'package:gql/ast.dart';
+
+/// returns true iff node has either
+/// - a @skip directive with an if-argument that evaluates to true
+/// - a @include directive with an if-argument that evaluates to false
+bool isSkipped(FieldNode node, Map<String, dynamic> variables) {
+  final directives = node.directives;
+
+  if (directives.isEmpty) {
+    return false;
+  }
+
+  final DirectiveNode? skipOrIncludeDirective;
+  final bool isSkipDirective;
+
+  final skipDirective = directives
+      .firstWhereOrNull((directive) => directive.name.value == 'skip');
+  if (skipDirective != null) {
+    skipOrIncludeDirective = skipDirective;
+    isSkipDirective = true;
+  } else {
+    final includeDirective = directives
+        .firstWhereOrNull((directive) => directive.name.value == 'include');
+    if (includeDirective != null) {
+      skipOrIncludeDirective = includeDirective;
+      isSkipDirective = false;
+    } else {
+      skipOrIncludeDirective = null;
+      isSkipDirective = false;
+    }
+  }
+
+  if (skipOrIncludeDirective == null) {
+    return false;
+  }
+
+  final ifArgument = skipOrIncludeDirective.arguments
+      .firstWhereOrNull((argument) => argument.name.value == 'if');
+
+  if (ifArgument == null) {
+    return false;
+  }
+
+  final ifValue = ifArgument.value;
+
+  // argument to if: is boolean literal -> use that literal directly
+  if (ifValue is BooleanValueNode) {
+    return isSkipDirective ? ifValue.value : !ifValue.value;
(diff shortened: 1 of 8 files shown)


knaeckeKami:
@budde377 @vincenzopalazzo

this might impact graphql and graphql codegen, so feel free to share your thoughts on this proposal.

see https://github.com/gql-dart/ferry/issues/432 and for possible adaptions to codegen, https://github.com/gql-dart/gql/pull/357/files#diff-1398a011e7a89cca7b15937b98e3f2baf1984b0225980cfa07cf5eb0cef97695R153

packages/normalize/lib/src/utils/well_known_directives.dart
```diff
@@ -0,0 +1,64 @@
...
+    if (variableValue is bool) {
+      return isSkipDirective ? variableValue : !variableValue;
+    }
+
+    /// TODO: throw error if variable is not a boolean?
+  }
+
+  /// TODO: throw error if ifValue is neither a boolean literal nor a variable?
+  return false;
+}
```
budde377, line 64:
```suggestion
import 'package:collection/collection.dart';
import 'package:gql/ast.dart';

/// returns true iff node has either
/// - a @skip directive with an if-argument that evaluates to true
/// - a @include directive with an if-argument that evaluates to false
bool isSkipped(FieldNode node, Map<String, dynamic> variables) {
  final directives = node.directives;

 
  final skipDirective = directives
      .firstWhereOrNull((directive) => const {'skip', 'include'}.contains( directive.name.value) );
  if (skipDirective == null) {
    return false;
  }
  final isSkipDirective = skipDirective.name.value === 'skip';

  final ifArgument = skipOrIncludeDirective.arguments
      .firstWhereOrNull((argument) => argument.name.value == 'if');

  if (ifArgument == null) {
    return false;
  }

  final ifValue = ifArgument.value;

  // argument to if: is boolean literal -> use that literal directly
  if (ifValue is BooleanValueNode) {
    return isSkipDirective ? ifValue.value : !ifValue.value;
  }
  // argument to if is a variable -> look up value of that variable in variables map
  if (ifValue is! VariableNode) {
    /// TODO: throw error if ifValue is neither a boolean literal nor a variable?
    return false;
  }
  final variableName = ifValue.name.value;
  final variableValue = variables[variableName];
  if (variableValue is! bool) {
   /// TODO: throw error if variable is not a boolean?
    return false;
  }
  return isSkipDirective ? variableValue : !variableValue;
}
```

budde377: A bit easier to read maybe? :)

budde377 (commented):
Great improvement! Should we also support fragments and inline fragments? 

It will not break `graphql_flutter` but definitely `graphql_codegen`, but I'll fix that.

packages/normalize/lib/src/utils/well_known_directives.dart
knaeckeKami: yes, thanks :)

knaeckeKami:
> Great improvement! Should we also support fragments and inline fragments?

Good point. Yes, it makes sense to support this consistently over fields, fragments and inline fragments. I'll add this later.

Issue #498, opened 2023-03-06
When evicting an entity from the cache, the list becomes empty when it is nested in an object.


ValentinVignal:
I have this code (you can also check out https://github.com/ValentinVignal/flutter_app_stable/tree/ferry/cache-and-fragment)

```graphql
# schema.graphql

type Pokemon {
  id: ID!
  name: String!
}

type Query {
  pokemons: [Pokemon!]!
  nestedPokemons: [NestedPokemon!]!
}

type NestedPokemon {
  nested: Pokemon!
}
```

With 2 queries: 

```graphql
# pokemons.graphql
query Pokemons {
  pokemons {
    id
    name
  }
}
```

```graphql
# nested_pokemons.graphql
query NestedPokemons {
  nestedPokemons {
    nested {
      id
      name
    }
  }
}
```

Notice that `pokemons` returns a list of `Pokemon` (with an id) while `nestedPokemon` returns a list of `NestedPokemon` pokemon (that doesn't have an id)

I have an application that:
1. Queries both lists (nested and non-nested)
2. Display sthem
3. Evicts one by one each entity

<details>

<summary>Code</summary>

```dart
import 'package:built_collection/built_collection.dart';
import 'package:ferry/ferry.dart';
import 'package:ferry_flutter/ferry_flutter.dart';
import 'package:flutter/material.dart';
import 'package:flutter_app_stable/__generated__/schema.ast.gql.dart';
import 'package:flutter_app_stable/graphql/__generated__/nested_pokemons.data.gql.dart';
import 'package:flutter_app_stable/graphql/__generated__/nested_pokemons.req.gql.dart';
import 'package:flutter_app_stable/graphql/__generated__/nested_pokemons.var.gql.dart';
import 'package:flutter_app_stable/graphql/__generated__/pokemons.data.gql.dart';
import 'package:flutter_app_stable/graphql/__generated__/pokemons.req.gql.dart';
import 'package:flutter_app_stable/graphql/__generated__/pokemons.var.gql.dart';
import 'package:gql_exec/src/response.dart';

final client = Client(
  link: Link.function((request, [forward]) {
    if (request.operation.operationName == 'Pokemons') {
      return Stream.value(
        Response(
          response: const {},
          data: GPokemonsData((data) {
            data.pokemons = ListBuilder(List.generate(
                40,
                (i) => GPokemonsData_pokemons(
                      (pokemon) => pokemon
                        ..id = '$i'
                        ..name = 'Pokemon $i',
[...]

PR #499, opened 2023-03-08
fix(normalize): Accept partial data in lists

ValentinVignal:
Fixes https://github.com/gql-dart/ferry/issues/498

### DIFF
--- a/examples/pokemon_explorer/test/dangling_reference_test.dart
+++ b/examples/pokemon_explorer/test/dangling_reference_test.dart
@@ -1,0 +1,98 @@
+import 'package:built_collection/built_collection.dart';
+import 'package:ferry/ferry.dart';
+import 'package:ferry_flutter/ferry_flutter.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+// ignore: depend_on_referenced_packages
+import 'package:gql_exec/src/response.dart';
+import 'package:pokemon_explorer/__generated__/schema.ast.gql.dart';
+import 'package:pokemon_explorer/__generated__/schema.schema.gql.dart';
+import 'package:pokemon_explorer/src/graphql/__generated__/all_pokemon.data.gql.dart';
+import 'package:pokemon_explorer/src/graphql/__generated__/all_pokemon.req.gql.dart';
+import 'package:pokemon_explorer/src/graphql/__generated__/all_pokemon.var.gql.dart';
+
+class _MyWidget extends StatelessWidget {
+  const _MyWidget({
+    required this.client,
+    Key? key,
+  }) : super(key: key);
+
+  final Client client;
+
+  @override
+  Widget build(BuildContext context) {
+    return MaterialApp(
+      home: Scaffold(
+        body: Operation<GAllPokemonData, GAllPokemonVars>(
+          client: client,
+          operationRequest: GAllPokemonReq(
+            (req) => req.vars
+              ..limit = 3
+              ..offset = 0,
+          ),
+          builder: (context, result, _) {
+            final pokemons = result?.data?.pokemons?.results ?? BuiltList();
+
+            return ListView.builder(
+              itemCount: pokemons.length,
+              itemBuilder: (context, index) {
+                final pokemon = pokemons[index];
+
+                return ListTile(
+                  title: Text(pokemon!.name),
+                );
+              },
+            );
+          },
+        ),
+      ),
+    );
+  }
+}
+
+void main() {
+  testWidgets("test description", (WidgetTester tester) async {
+    final client = Client(
+      cache: Cache(possibleTypes: possibleTypesMap),
+      link: Link.function((request, [forward]) {
+        return Stream.value(
+          Response(
+            response: const {},
+            data: GAllPokemonData((data) {
+              data.pokemons.results = ListBuilder(List.generate(
+                  3,
+                  (i) => GAllPokemonData_pokemons_results(
+                        (pokemon) {
+                          pokemon
+                            ..id = i
+                            ..name = 'name$i'
+                            ..avatar = 'avatar$i'
+                            ..height.in_meter = 'height$i'
+                            ..weight.in_kg = 'weight$i';
+                        },
+                      )));
+            }).toJson(),
+          ),
+        );
+      }),
+    );
+
+    await tester.pumpWidget(_MyWidget(client: client));
+    await tester.pump();
+    expect(find.byType(ListTile), findsNWidgets(3));
+    for (final id in List.generate(3, (index) => index)) {
+      expect(find.widgetWithText(ListTile, 'name$id'), findsOneWidget);
+    }
+
+    client.cache
+      ..evict('${Pokemon.name.value}:1')
+      ..gc();
+    await tester.runAsync(() => Future.delayed(Duration.zero));
+    await tester.pump();
+
+    expect(find.byType(ListTile), findsNWidgets(2));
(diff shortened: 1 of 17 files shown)


ValentinVignal:
https://user-images.githubusercontent.com/32538273/223689497-3564aff7-22de-4eb4-999a-a6d7d32b8989.mov

It fixes the example given in https://github.com/gql-dart/ferry/issues/498

knaeckeKami:
Thanks for the PR! 

I know I suggested this change myself, but I'm afraid I did not fully think this through.

I can see some some issues with this.

Example:

Let's say there's a query 

```graphql
query Pokemons {
  pokemons {
    id
    name
  }
}
```

which is executed with .CacheFirst FetchPolicy.
Then, another field is added to the query, so now it's:

```graphql
query Pokemons {
  pokemons {
    id
    name
    weight
  }
}
```

In the current release, this would cause the query to be re-executed, since the missing weight fields in the cache cause a partial data exception.

With this change, this would just drop all the items from the list and it would return an empty list, but would not actually re-execute it on the network.

ValentinVignal:
> I'm afraid the real fix for this is not that easy. We should probably not check for any partial data, but really just for dangling references in child objects of lists.

What do you mean by that? That today's behavior of ferry is what we should keep?

Or I could try to add some parameters to `denormalizeNode` like `allowPartialInList` or create another exception `MissingDanglingReference` ?

knaeckeKami:
I think we need to distinguish between two cases: 
real partial data (like missing fields) -> keep the behaviour now and return null / throw a partial data exception (depending on config) so that clients can now if they need to execute another network request or if cached data is good.

dangling references in nested child objects -> In this case, I think it's fine if we remove child objects, since they were likely manually removed from the cache. I still think it should be configurable in order to not break existing users with this change. 

So yes, I think additional config parameters or exceptions are the way to go.

ValentinVignal:
@knaeckeKami I'm willing to give it a try if you give me some time :)

ValentinVignal:
@knaeckeKami I pushed some new codes which adds `allowDanglingReference` to the config. Tell me if you want me to do something more

ValentinVignal:
It looks like there is an issue with melos configuration, the CI is failing with 

```
Run melos bs
  melos bs
  shell: /usr/bin/bash -e {0}
  env:
    FLUTTER_ROOT: /opt/hostedtoolcache/flutter/beta-3.9.0-0.[2](https://github.com/gql-dart/ferry/actions/runs/4649666127/jobs/8228167917?pr=499#step:5:2).pre-x64
    PUB_CACHE: /opt/hostedtoolcache/flutter/beta-[3](https://github.com/gql-dart/ferry/actions/runs/4649666127/jobs/8228167917?pr=499#step:5:3).9.0-0.2.pre-x6[4](https://github.com/gql-dart/ferry/actions/runs/4649666127/jobs/8228167917?pr=499#step:5:4)/.pub-cache
Found a melos.yaml file in "/home/runner/work/ferry/ferry" but no local installation of Melos.

From version 3.0.0, the melos package must be installed in a pubspec.yaml file next to the melos.yaml file.

For more information on migrating to version 3.0.0, see: https://melos.invertase.dev/guides/migrations#200-to-300

To migrate at a later time, ensure you have version 2.[9](https://github.com/gql-dart/ferry/actions/runs/4649666127/jobs/8228167917?pr=499#step:5:10).0 or below installed: dart pub global activate melos 2.9.0
Error: Process completed with exit code 1.
```


I created a PR to set the version to `2.9.0` https://github.com/gql-dart/ferry/pull/506

examples/pokemon_explorer/test/dangling_reference_test.dart
```diff
@@ -0,0 +1,98 @@
+import 'package:built_collection/built_collection.dart';
+import 'package:ferry/ferry.dart';
+import 'package:ferry_flutter/ferry_flutter.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+// ignore: depend_on_referenced_packages
+import 'package:gql_exec/src/response.dart';
```
ValentinVignal, line 7: Do you prefer me to add `gql_exec` to the dev dependencies of the example?

knaeckeKami: yes, please

packages/normalize/lib/src/denormalize_node.dart
```diff
@@ -19,24 +18,33 @@ Object? denormalizeNode({
...
+    return newList;
   }
 
   // If this is a leaf node, return the data
   if (selectionSet == null) return dataForNode;
 
   if (dataForNode is Map) {
-    final denormalizedData = dataForNode.containsKey(config.referenceKey)
-        ? config.read(dataForNode[config.referenceKey]) ?? {}
+    final isReference = dataForNode.containsKey(config.referenceKey);
```
knaeckeKami, line 45: I think you should just add the check for dangling references below this line.

if it's a reference, and the reference is dangling, and dangling references are allowed, throw a danglingReferenceException.

After the reference has been de-referenced successfully and the data is check for partial data, the behaviour should not change, I think

knaeckeKami:
Thanks for the update.

I think I spotted a change in behaviour; I added an integration test in test/client/partial_query_data_test.dart. 

This tests the following scenario:

- the client fetches a query and caches the result
- in a newer version of the app, a new field is added to that query
- the client executes the updated version of the query with FetchPolicy.CacheFirst
- since the cached data is partial, the cache should return no data, and the terminal network link should execute the query

This tests succeeds in the current version on master, but I believe it fails with your changes. 

I added a comment on where I think the issue is.

ValentinVignal:
Thank you for the review @knaeckeKami ! I updated the code in [bug: Only throw dangling reference exception when it is one](https://github.com/gql-dart/ferry/pull/499/commits/fbb3c58cb66dc950c14f5c280e1c937e8850fdee) to make your test pass

ValentinVignal:
@knaeckeKami Is there anything else I should do on this PR ?

knaeckeKami:
Sorry, I did not have time to check the updates, will do shortly

packages/normalize/lib/src/normalize_fragment.dart
```diff
@@ -40,6 +39,7 @@ void normalizeFragment({
   bool addTypename = false,
   String referenceKey = kDefaultReferenceKey,
   bool acceptPartialData = true,
+  bool acceptDanglingReferences = false,
```
knaeckeKami, line 42: does this parameter do anything when normalizing a fragment?

if not, please remove it -
if yes - please rename it to allowDanglingReferences for consistence

packages/normalize/lib/src/normalize_operation.dart
```diff
@@ -35,6 +34,7 @@ void normalizeOperation({
   DataIdResolver? dataIdFromObject,
   bool addTypename = false,
   bool acceptPartialData = true,
+  bool acceptDanglingReferences = false,
```
knaeckeKami, line 37: same as above

knaeckeKami (changes requested):

packages/normalize/lib/src/normalize_fragment.dart
ValentinVignal: Oops, sorry for that, it doesn't do anything. I deleted it in [refactor: Remove unused parameter for normalize methods](https://github.com/gql-dart/ferry/pull/499/commits/f36583c3731613fdddfa755805a4bf7a2a935b24)

knaeckeKami:
LGTM!

PR #530, opened 2023-08-13
feat(ferry_generator): add option to reuse data classes for fragments with a single inline fragment  spread as selection

knaeckeKami:
- chore: update codegen
- feat(ferry_generator): add data_class_config: reuse_fragment option in order to re-use selection sets with only a single inline fragment spread. see https://github.com/gql-dart/gql/pull/413 for details.


knaeckeKami:
cc @dehypnosis

Issue #531, opened 2023-08-16
[FerryGenerator]How to make it work within a monorepo


letsar:
Hi!

First of all, thanks for this suite of packages.
We are using Ferry within a monorepo and using `ferry_generator` to generate the models. We would want to have each graphQL query in the dedicated package, it works well, but we need to have the schema in each packages and thus, all the generated code for the schema is duplicated in each package.

We would want to have the schema in a dedicated package and be able to only generate the models of the schema in this package and not in the other ones.

Did I miss something or it's not possible for the moment?

knaeckeKami:
Well, actually it's not that simple.
Currently there is the assumption that there is a `Serializers` object that knows how to deserialize any OperationRequest without type hints (i.e. for the offline mutation link), this might break with a multi module approach.

letsar:
Hi thanks for taking the time to look at this.
I don't have the time right now to make a POC, but maybe in a few weeks. I have a few ideas. Let me know what you think.

- One package would be responsible for generating the models for the schema and the serializers.
- I think we would still need to have the schema in every package (or at list a symbolic link) for the `graphql_builder` to work.
- We could have an option in the `build.yaml` file to specify the imports for the schema models and the serializers, so that we could use them in the generated code for the queries. If null, we would consider that the schema models are in the same package.
- Maybe we could remove the `serializer_builder` from the build.yaml for the packages which don't generate the models for the schema.

knaeckeKami:
A single package for the schema & the serializers would not work, I fear.

If we have a single package for the schema, and N feature packages that contain operations (queries, mutations, subscriptions), the feature packages need to depend on the schema.

But the serializers need to know both the schema and the operations, so this would cause circular dependencies.

I think a way to solve this would be:

- have a schema package that every feature package depends on. The schema package generates the .schema.gql.dart file and a Serializers class, but just for the schema types (e.g. scalars, enums, inputs). 

- feature packages depend on the schema file and import the generated .schema.gql.dart file instead of generating their own. They also generate their own `Serializers` class and merge it with the `Serializers` (https://pub.dev/documentation/built_value/latest/serializer/Serializers/merge.html ) from the schema in order so ensure that they can fully serialize all required types.

- optional: when required (e.g. for persisting mutations when offline), add an additional package that depends on the schema package and all feature packages. This package merges all the Serializers so we can again serialize and deserialize all types with a single object. 


This might sound pretty scary, but I think all of this is doable with relatively small changes to the code.

[...]

PR #533, opened 2023-08-16
feat: allow importing schema from other package

knaeckeKami:
experiment for supporting monorepos


meg4cyberc4t:
Don't give it up! A very important feature!

MattisBrizard:
Hi @knaeckeKami, I think this would be a top feature for monorepos 
Do you think that there is still much to do to ship this feature ?

knaeckeKami:
IMO it won't take much more to ship this, but I kinda abandoned it since there is an issue:

The current architecture uses built_value, it generates a `Serializers` object which knows how to serialize each generated built_value type.

ferry currently assumes that this serializers type can serialize all types - at least for offline mutations that's required.

But in a monorepo, every repo would generate its own `Serializers` instance, which only knows how to serialize the types form this repo.

I don't have a good solution for this right now.

MattisBrizard:
It's not possible to have each package generating their own `Serializers` and merge it with the as you mentioned here :
> * feature packages depend on the schema file and import the generated .schema.gql.dart file instead of generating their own. They also generate their own `Serializers` class and merge it with the `Serializers` ([pub.dev/documentation/built_value/latest/serializer/Serializers/merge.html](https://pub.dev/documentation/built_value/latest/serializer/Serializers/merge.html) ) from the schema in order so ensure that they can fully serialize all required types.

MattisBrizard:
It's not possible to have each package generate its own Serializers and merge it with the one from the schema?
As you mentioned [here](https://github.com/gql-dart/ferry/issues/531#issuecomment-1681980509):

> * feature packages depend on the schema file and import the generated .schema.gql.dart file instead of generating their own. They also generate their own `Serializers` class and merge it with the `Serializers` ([pub.dev/documentation/built_value/latest/serializer/Serializers/merge.html](https://pub.dev/documentation/built_value/latest/serializer/Serializers/merge.html) ) from the schema in order so ensure that they can fully serialize all required types.

knaeckeKami:
That might be possible, but that would introduce cyclic dependencies; the final `Serializers ` would need to depend on every  repository in the monorepo.
And then, what's the point of having a monorepo?

knaeckeKami:
fixed by #606

Issue #575, opened 2024-02-02
[Feature Request] Add `ContextEntries` to requests that can be retrieved in custom `gql` `Link`


tpucci:
In order to create advanced `Link`s that can execute custom logic per request, we need a way to extend request with `ContextEntries`.

Currently, generated code for a request is:

```dart
  @override
  Request get execRequest => Request(
    operation: operation,
    variables: vars.toJson(),
    // no passed context. Thus, the default is `context = const Context()`
  );
```

`gql_exec` has a method `.withContextEntry` on `Operation` that we could use.

I'd be happy to contribute if you think it is a good issue 🙂

_(😅 Hope I did not miss something in the API that already allows us to create context entries)_
_Reference:_ #556

tpucci:
Hello @knaeckeKami !
Alright, here is my plan. What do you think ?

1. Extend `OperationRequest` with a `withContextEntry<T extends ContextEntry>(T entry)` method
2. Modify codegen to implement `withContextEntry` and pass it to `execRequest`

knaeckeKami:
One way would be to add a `Context?` getter to the abstract OperationRequest class and adapt ferry_codegen to add this to the generated Request classes.

This would already allow users to add the context and modify Contexts via the built_value builders;

`withContextEntry()` is probably not strictly necessary since the built value classes already support that in their rebuild() methods, but it would allow to work with the context in TypedLinks where the exact type is not known.

A challenge is the serialization of the OperationRequest; Since the context is not serializable (it's a map of <Type, ContextEntry>, we would have to exclude it like `updateResult`.

tpucci:
Ok got it. So:

1. `Context?` getter
2. codegen & exclude from serialization

knaeckeKami:
sounds good to me!

bawahakim:
@tpucci This is a life saver, thank you! Would it be possible to also have a `withContextEntry` and `updateContextEntry` in the `OperationRequest` so we can also modify the context globally, but before any link? My use case is I need to add a unique transaction id for each request header (not the same as the `requestId`), and also want to pass the original stack trace of the caller to simplify debugging (stack trace in chains loses the original caller info it seems).

tpucci:
Hey @bawahakim 
I am not sure I understand.
`withEntry` and `updateEntry` exist on `Context`, you could use those methods 🙂

If you want to set your context per request:

```dart
client.request(GSomeOperation((b) {
        b
          ..context = const Context().withEntry(YourEntry());
[...]

Issue #556, opened 2023-12-02
[Question] Is there a way to pass specific headers per request?


Pacane:
Hi,

I'd like to be able to pass specific headers on a per request base. I know I can do this by re-creating a client using another `HttpLink` with the headers I want to pass, but is there a way to do this with an existing client?

I used to use Artemis and we could pass in the `Context` on `execute`, which allowed me to override the headers in the request. Is there such a feature within Ferry?

tpucci:
I think it can not be done because ferry generated code is:

```dart
  @override
  Request get execRequest => Request(
    operation: operation,
    variables: vars.toJson(),
  );
```

...which can not be passed a `context` 😅 I'll opened a feature request.

PR #576, opened 2024-02-03
feat(ferry_exec): Pass context to GQL Links

tpucci:
- feat: pass context to gql links
- feat: update generated code

> Generated code was updated with "void" statements. I am not sure why

Feature request: #575 
> In order to create advanced `Link`s that can execute custom logic per request, we need a way to extend request with `ContextEntries`.
> 
> Currently, generated code for a request is:
> 
> ```dart
>   @override
>   Request get execRequest => Request(
>     operation: operation,
>     variables: vars.toJson(),
>     // no passed context. Thus, the default is `context = const Context()`
>   );
> ```
> 
> `gql_exec` has a method `.withContextEntry` on `Operation` that we could use.
> 
> I'd be happy to contribute if you think it is a good issue 🙂
> 
> _(😅 Hope I did not miss something in the API that already allows us to create context entries)_ _Reference:_ #556

### DIFF
--- a/examples/pokemon_explorer/lib/src/graphql/__generated__/all_pokemon.data.gql.dart
+++ b/examples/pokemon_explorer/lib/src/graphql/__generated__/all_pokemon.data.gql.dart
@@ -15,7 +15,7 @@ abstract class GAllPokemonData
     implements Built<GAllPokemonData, GAllPokemonDataBuilder> {
   GAllPokemonData._();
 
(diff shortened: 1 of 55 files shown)


tpucci:
I'd be happy to fix `verify_commits` but I did not understand why it failed

knaeckeKami:
It was the name of the PR. edited it for you and reran the action :)

(this is so CHANGELOGs can be created automatically with melos)

knaeckeKami:
LGTM!

Issue #582, opened 2024-02-15
Generating tristate_optionals generates invalid code with mixed up imports


thomas7D:
I need to use tristate optionals for my mutations and ran into invalid code generation when enabling it.

I recreated the issue in a [simple project](https://github.com/thomas7D/ferry_tristate_bug).

The generated `test.var.gql.dart` contains errors as it's mixing up the imports when using the tristate classes.

```
import 'package:ferry_test/__generated__/serializers.gql.dart' as _i1;
import 'package:gql_tristate_value/gql_tristate_value.dart' as _i2;
...
      b..name = const _i2.AbsentValue();
...
if (_$namevalue case _i1.PresentValue(value: final _$value)) {
```

In some lines it uses the correct import `_i2` for the tristate value, in other it's wrong leading to syntax errors.

I haven't figured out what causes this issue as it's not present in all cases.
In the example project you can comment out the `Pokemons` query in `lib/test.graphql` and the code generation works as expected. But when both query and mutation are present the code is broken.

The same problems also occured in the generated schema files in a larger project when modifying that, with no obvious correlation between the changes in the schema and the mixing of imports.

knaeckeKami:
Do you also have multiple operations in the same file in your larger project?

thomas7D:
Yes, I group related queries and mutations for the same objects in a single file.

knaeckeKami:
Thanks! can you reproduce this issue when you put all operations in a separate file?

thomas7D:
No, code is fine when I put the mutation in a separate graphql file.

I tested different variations of putting operations in single files, grouping them slightly, etc. and stumbled on this:

If I put both operations in the same file like this, I get the wrong imports prefixes:
```
query Pokemons {
    pokemons {
        id
        name
    }
}

mutation UpdatePokemon($id: ID!, $name: String) {
    updatePokemon(id: $id, name: $name) {
        id
        name
    }
}
```

If I switch them around and rerun build_runner, I get no errors in the generated files:
```
mutation UpdatePokemon($id: ID!, $name: String) {
    updatePokemon(id: $id, name: $name) {
        id
        name
    }
}

query Pokemons {
    pokemons {
        id
        name
    }
}
```

So I don't think it's related to grouping them or spreading them into separate files.

LiLatee:
Ok, I think I found a solution with the use `typeOverrides` map, but I need to check it. I will back to you

LiLatee:
Hm I am a little bit blocked.
[...]

PR #583, opened 2024-02-19
#fix: allocator for tristate_optional usage

LiLatee:
Fix for the issue https://github.com/gql-dart/ferry/issues/582.

Explanation:
- used allocators had always 0 imports, because `allocate` method was never called.

@knaeckeKami I am not familiar with the tests there. Could you explain how to fix them? 🙏 

## TODO:
- fix test


knaeckeKami:
I am also not familar with these tests, but given the failure i highly suspect that the import prefix changed due to your changes

PR #587, opened 2024-04-04
feat(ferry): allow sharing of clients between Isolates

odlund:
* Makes the commandSendPort visible outside of the package
* Create a new `IsolateClient.attach(SendPort)` function to attach to an existing client from a separate Isolate


knaeckeKami:
Thanks! It seems the pipeline failures are due to new upgrades of the underlying SDKs and unrelated to the changes, no action item needed from your side.

PR #614, opened 2024-10-22
refactor(normalize,ferry,ferry_generator): Add const constructor to visitors

ValentinVignal:
This adds a const constructor to the visitors.

### DIFF
--- a/packages/normalize/lib/src/utils/well_known_directives.dart
+++ b/packages/normalize/lib/src/utils/well_known_directives.dart
@@ -5,7 +5,7 @@ import 'package:gql/ast.dart';
 /// - a @skip directive with an if-argument that evaluates to true
 /// - a @include directive with an if-argument that evaluates to false
 bool isSkipped(SelectionNode node, Map<String, dynamic> variables) {
-  final directives = node.accept(_SelectionNodeDirectiveVisitor.instance);
+  final directives = node.accept(const _SelectionNodeDirectiveVisitor());
 
   final skipOrIncludeDirective = directives?.firstWhereOrNull(
       (directive) => const {'skip', 'include'}.contains(directive.name.value));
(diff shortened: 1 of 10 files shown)


ValentinVignal:
Do I need to make different pull requests for the different packages?

packages/normalize/lib/src/utils/well_known_directives.dart
```diff
@@ -41,7 +41,9 @@ bool isSkipped(SelectionNode node, Map<String, dynamic> variables) {
 
 class _SelectionNodeDirectiveVisitor
     extends SimpleVisitor<List<DirectiveNode>?> {
-  static final instance = _SelectionNodeDirectiveVisitor();
+  const _SelectionNodeDirectiveVisitor();
+
+  static const instance = _SelectionNodeDirectiveVisitor();
```
knaeckeKami, line 46: since the class is now const, we can remove the static instance and replace it with `_SelectionNodeDirectiveVisitor()`

knaeckeKami:
> Do I need to make different pull requests for the different packages?

for refactors like this I'm ok if it affects multiple packages, just make sure to use conventional commits with the package names so automatic versioning works

ValentinVignal:
> for refactors like this I'm ok if it affects multiple packages, just make sure to use conventional commits with the package names so automatic versioning works

Sure let me do that!

ValentinVignal:
> for refactors like this I'm ok if it affects multiple packages, just make sure to use conventional commits with the package names so automatic versioning works

@knaeckeKami, I updated the commits and the title of the PR. Is it what you were expecting?

packages/normalize/lib/src/utils/well_known_directives.dart
ValentinVignal: You're right, I included it in [refactor(normalize): Add const constructor to visitors](https://github.com/gql-dart/ferry/pull/614/commits/c7474585ab49bdcd2744880da7527fdfd51988f2)

knaeckeKami (approved):

knaeckeKami:
Thanks1

PR #629, opened 2025-02-13
RequestControllerTypedLink return subscription streams that can end

warrenisarobot:
Update RequestControllerTypedLink so that it will not refetch and return the stream result directly.  This allows the stream to end if the server completes the stream, and the listeners `onData` will fire.

Since the requestController stream never closes the switchMap's result stream will also never reach the onDone state.  This uses only the first matching requestController emitted request when it is a subscription operation.  This allows the switchMap's resulting stream to close when the server's response stream closes as well.

### DIFF
--- a/packages/ferry/lib/src/request_controller_typed_link.dart
+++ b/packages/ferry/lib/src/request_controller_typed_link.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 
-import 'package:ferry_exec/ferry_exec.dart';
+import 'package:ferry/ferry.dart';
+import 'package:gql_exec/gql_exec.dart';
 import 'package:rxdart/rxdart.dart';
 
 /// Allows multiple requests to be made by adding requests to the
@@ -57,7 +58,7 @@ class RequestControllerTypedLink extends TypedLink {
       // If no stream has been cached for this request, create a new one.
       ValueStream<OperationResponse<TData, TVars>>? prev;
       var initial = true;
-      stream = requestController.stream
+      var requestStream = requestController.stream
           .whereType<OperationRequest<TData, TVars>>()
           .where(
             (req) => req.requestId == null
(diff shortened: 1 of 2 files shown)


knaeckeKami:
Hi!

Please add a test that ensures that this new behavior will not break in the future.

should be something like

```dart
  group('done event for subscriptions', () {
    late TypedLink typedLink;

    setUp(() {
      typedLink = TypedLink.from([
        RequestControllerTypedLink(),
        TypedLink.function(<TData, TVars>(request, [next]) =>
            Stream<OperationResponse<TData, TVars>>.empty()),
      ]);
    });


    test('onDone is propagated for subscriptions', () {
      final stream = typedLink.request(
        JsonOperationRequest(
          fetchPolicy: FetchPolicy.NetworkOnly,
          vars: {},
          operation: Operation(
            document: gql.parseString(r'''
            subscription Sub {
                reviews {
                  id
                  stars
                }
            }'''),
          ),
        ),
      );

      expect(stream, emitsInOrder([emitsDone]));
    });
  });
```

packages/ferry/lib/src/request_controller_typed_link.dart
```diff
@@ -71,7 +72,11 @@ class RequestControllerTypedLink extends TypedLink {
           final sub = prev?.listen(null);
           Future.delayed(Duration.zero, () => sub?.cancel());
         },
-      ).switchMap(
+      );
+      if (request.operation.getOperationType() == OperationType.subscription) {
+        requestStream = requestStream.take(1);
```
knaeckeKami, line 77: Clever solution!

but I'm afraid that if I look at this in a few months, I won't remember on why this is necessary.

Please add a comment like

```
Only use the first instance of the request stream for subscriptions, to ensure that a possible .done event
is propagated.
Note: This disables the paging and refetch feature for subscriptions.
```

warrenisarobot: Great point, added comments on this

knaeckeKami:
Thanks!

PR #640, opened 2025-05-23
feat(ferry_generator): add option to disable formatting

Vild:
Adds a flag to the build.xml's options to prefix the generated code with `// dart format off`.
I set the default value is true to not affect any existing output.

Do I need to make any other changes for this simple feature?


knaeckeKami:
Thanks and sorry for the long delay, i missed this.

Please document this feature in

docs/codegen.md, this will be used the generate the website.

